PR #2, opened 2017-07-07
[#146287083] Mary wants to know when an event is full

grabartley:


src/events/cd-event-list.vue
```diff
@@ -32,14 +32,21 @@
       </div>
       <div class="cd-event-list__event-view-wrapper">
         <router-link :to="{name: 'EventDobVerification', params: {eventId: event.id}}"
+                     v-if="!isEventFull(event.id)"
                      tag="button" class="btn btn-lg btn-primary cd-event-list__event-view">
           {{ $t('See Details and Book') }}
         </router-link>
+        <router-link :to="{name: 'EventDobVerification', params: {eventId: event.id}}"
```
DanielBrierton: Since this is disabled, should probably just use a regular `<button>` tag instead, since the `router-link` will never be able to go anywhere anyway.

DanielBrierton: Alternatively, you could have just one element, and use `isEventFull` to decide whether it's disabled, and what text to display.

src/events/cd-event-list.vue
```diff
@@ -32,14 +32,21 @@
...
+                     v-if="isEventFull(event.id)" disabled="disabled"
+                     tag="button" class="btn btn-lg cd-event-list__event-view cd-event-list__event-view--disabled">
```
DanielBrierton: No need for the `--disabled` modifier. We can use the [:disabled pseudo selector](https://developer.mozilla.org/en-US/docs/Web/CSS/:disabled) instead.

src/events/cd-event-list.vue
```diff
@@ -32,14 +32,21 @@
...
+          {{ $t('FULL') }}
```
DanielBrierton: Should use `Full` instead, and use `text-transform: uppercase;` to make it all caps instead. Makes it more reusable for other situations where we don't want it all caps.

src/events/cd-event-list.vue
```diff
@@ -65,6 +72,18 @@
       getSessionListForEvent(event) {
         return event.sessions.map(session => session.name).join(', ');
       },
+      isEventFull(eventId) {
+        const currentEvent = this.events.find(event => event.id === eventId);
+        let totalEventCapacity = 0;
+        let totalTicketsBooked = 0;
+        forEach(currentEvent.sessions, (session) => {
```
DanielBrierton: Probably better off just using the native `forEach` function of arrays instead of lodash.

```javascript
currentEvent.sessions.forEach((session) => {
  session.tickets.forEach((session) => {
    totalEventCapacity += ticket.quantity;
    totalTicketsBooked += ticket.approvedApplications;
  });
});
```

The less unnecessary dependencies, the better! 😅

test/e2e/specs/book-event.js
```diff
@@ -438,4 +438,11 @@ describe('Book event page', () => {
     Booking.modifyButton.click();
     EventSessionsPage.sessionsHeader.waitForVisible();
   });
+
+  it('should not allow the user to continue in the booking flow when the event is full', () => {
+    DojoPage.openDojoWithLatLong(10, 89, 1);
+    DojoDetailsPage.name.waitForVisible();
+    expect(DojoDetailsPage.eventViewButtons(0).getText()).equals('FULL');
+    expect(DojoDetailsPage.eventViewButtons(0).isEnabled()).equals(false);
```
DanielBrierton: Might be good to click the button and assert that we haven't changed page. The `isEnabled()` should be enough though...

src/events/cd-event-list.vue
```diff
@@ -65,6 +72,18 @@
       getSessionListForEvent(event) {
         return event.sessions.map(session => session.name).join(', ');
       },
+      isEventFull(eventId) {
+        const currentEvent = this.events.find(event => event.id === eventId);
```
DanielBrierton: You can skip this step by just passing the whole `event` instead of the `eventId`.

PR #6, opened 2017-07-18
[#148849175] Mary wants to navigate to the Dojos website directly from their listing.

grabartley:


src/dojos/cd-dojo-details.vue
```diff
@@ -21,7 +21,7 @@
           <a :href="'mailto:' + dojoDetails.email">{{ dojoDetails.email }}</a>
         </info-column-section>
         <info-column-section class="hidden-xs" icon="globe" :header="$t('Website')">
-          {{ dojoDetails.website }}
+          <a :href="'http://' + dojoDetails.website" target="_blank">{{ dojoDetails.website }}</a>
```
DanielBrierton: The issue with this story is that `dojoDetails.website` could actually have `http://` at the start already. Also, could be https rather than http.

PR #8, opened 2017-07-20
[#147740351] Implemented OpenGraph tags for dojos

DanielBrierton:
(Yes, I'm aware I accidentally pushed this branch to CoderDojo instead of my own fork...)

### DIFF
--- a/index.ejs
+++ b/index.ejs
@@ -4,9 +4,17 @@
     <meta charset="utf-8">
     <title>cp-zen-frontend</title>
     <meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1, user-scalable=no">
+    <% if (htmlWebpackPlugin.options.openGraph) { %>
+      {{#each openGraphProperties}}
+        {{#each this}}
+          <meta name="og:{{@../key}}" content="{{this}}">
+        {{else}}
+          <meta name="og:{{@key}}" content="{{this}}">
+        {{/each}}
+      {{/each}}
+    <% } %>
     <link rel="stylesheet" href="//fonts.googleapis.com/css?family=Lato:300,400,700" type="text/css" />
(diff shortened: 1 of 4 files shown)


index.ejs
```diff
@@ -4,9 +4,17 @@
...
+        {{#each this}}
+          <meta name="og:{{@../key}}" content="{{this}}">
+        {{else}}
+          <meta name="og:{{@key}}" content="{{this}}">
+        {{/each}}
+      {{/each}}
+    <% } %>
     <link rel="stylesheet" href="//fonts.googleapis.com/css?family=Lato:300,400,700" type="text/css" />
     <script src="https://www.google.com/recaptcha/api.js?onload=vueRecaptchaApiLoaded&render=explicit" async defer></script>
-    <link rel="stylesheet" href="./assets/style.css" type="text/css" />
```
Wardormeur: Why was this removed?

index.js
```diff
@@ -28,10 +33,17 @@ exports.register = function (server, options, next) {
 
   server.route({
     method: 'GET',
-    path: '/dojos/{country}/{path*}',
-    handler: {
-      file: {
-        path: __dirname + '/dist/index.html'
+    path: '/dojos/{id}/{alpha2*}',
```
Wardormeur: path changed, does that mean the default page url for dojos is also changed?

index.ejs
DanielBrierton: The file doesn't exist, and has been resulting in a 404 since the start of the project 😅  It also caused issues with OpenGraph as it was a relative URL, and so Hapi picked it up as a Dojo URL

index.js
Wardormeur: Nvmind, it's to match the preloader values am i right ?

DanielBrierton: The only thing that's looking at the names of these params is the dojo preloader, so I just changed it to match what it expects.

DanielBrierton: (So, basically, yes 😛 )

PR #9, opened 2017-07-21
[#148129617] Mary wants to avoid booking for a private dojo when she is not a member of the dojo

grabartley:


src/dojos/service.js
```diff
@@ -4,9 +4,18 @@ import GeolocationService from '@/geolocation/service';
 const radius = 50000;
 
 const DojosService = {
+  getDojoById: id => Vue.http.get(`${Vue.config.apiServer}/api/2.0/dojos/${id}`),
+
   getDojos: query => Vue.http.post(`${Vue.config.apiServer}/api/2.0/dojos`, { query }),
 
-  getDojoById: id => Vue.http.get(`${Vue.config.apiServer}/api/2.0/dojos/${id}`),
+  isUserMemberOfDojo(userId, dojoId) {
```
DanielBrierton: The method starting with `is` suggests a boolean being returned. Should be just `getUsersDojos` as that's what it really does.

json-server/server.js
```diff
@@ -103,6 +105,21 @@ server.get('/api/2.0/users/instance', (req, res) => {
...
 });
+
+server.post('/api/2.0/dojos/users', (req, res) => {
+  let dojos;
+  if (usersDojos[req.body.query.userId]) {
+    dojos = usersDojos[req.body.query.userId];
+    if (dojos[req.body.query.dojoId]) {
+      res.send(dojos[req.body.query.dojoId]);
+    } else {
+      res.send(null);
```
DanielBrierton: I think this should return `[]` if no match is found. Same below on line 119

src/events/cd-event-list.vue
```diff
@@ -91,6 +93,15 @@
         return totalEventCapacity === totalTicketsBooked;
       },
     },
+    watch: {
+      currentUser(newUser) {
+        if (newUser) {
+          DojosService.isUserMemberOfDojo(newUser.id, this.dojo.id).then((response) => {
+            this.isMember = !!response.body;
```
DanielBrierton: Would be a bit cleaner to assign body to `this.usersDojos` or something like that, then have a computed property for `isMember`. Allows us to use the data in other ways in the future, like showing admin controls to admins.

PR #11, opened 2017-07-24
[#149098899] Mary wants to see a Dojos eventbrite events

grabartley:


src/events/cd-event-list.vue
```diff
@@ -31,8 +31,11 @@
         </div>
       </div>
       <div v-if="canBook" class="cd-event-list__event-view-wrapper">
+        <div v-if="event.eventbriteId">
+          <a :href="event.eventbriteUrl | cdUrlFormatter" target="_blank" class="btn btn-lg btn-primary cd-event-list__event-view">{{ $t('See Details and Book') }}</a>
+        </div>
         <router-link :to="{name: 'EventDobVerification', params: {eventId: event.id}}"
```
DanielBrierton: Really minor, but you could use `v-else` here instead of another `v-if`.

https://vuejs.org/v2/guide/conditional.html#v-else

src/common/filters/cd-url-formatter.js
```diff
@@ -0,0 +1,6 @@
+export default (url) => {
+  if (url.substring(0, 7) === 'http://' || url.substring(0, 8) === 'https://') {
+    return url;
+  }
+  return `http://${url}`;
```
Wardormeur: https by default rather than http ?

test/unit/specs/common/filters/cd-url-formatter.spec.js
```diff
@@ -0,0 +1,10 @@
+import cdUrlFormatter from '@/common/filters/cd-url-formatter';
+
+describe('Url formatting filter', () => {
+  it('should format url to include http:// or https://', () => {
+    expect(cdUrlFormatter('www.example.com')).to.equal('http://www.example.com');
+    expect(cdUrlFormatter('http://www.example.com')).to.equal('http://www.example.com');
+    expect(cdUrlFormatter('https://www.example.com')).to.equal('https://www.example.com');
+    expect(cdUrlFormatter('http:/www.example.com')).to.equal('http://http:/www.example.com');
```
Wardormeur: uh.. what ?

src/common/filters/cd-url-formatter.js
DanielBrierton: This is for external sites. We don't know if they support https or not. I'd imagine most Dojo sites wouldn't. Any that do would usually redirect from http to https anyway, or even have HSTS so it won't even hit their server over http.

Wardormeur: GET/HEAD === 200 ? :/

test/unit/specs/common/filters/cd-url-formatter.spec.js
DanielBrierton: I guess it's a slightly pointless test, but I get the idea behind it. It has to be a full valid URL or it gets prefixed. Really this tests the same as the first test, but I guess this additionally checks that we're not being really naïve and checking that the string starts with 'http'

src/common/filters/cd-url-formatter.js
DanielBrierton: CORS

grabartley: CORS ?

test/unit/specs/common/filters/cd-url-formatter.spec.js
grabartley: Can remove it?

src/common/filters/cd-url-formatter.js
Wardormeur: Nobody puts cors on HEAD :D

test/unit/specs/common/filters/cd-url-formatter.spec.js
Wardormeur: Well, my point was more about fixing the filter instead  :/

src/events/cd-event-list.vue
grabartley: Will update this, thanks :)

test/unit/specs/common/filters/cd-url-formatter.spec.js
DanielBrierton: What's to be fixed? It shouldn't figure out that there's a `/` missing if you ask me.

src/common/filters/cd-url-formatter.js
DanielBrierton: No custom setup on my server:
https://codepen.io/danielbrierton/pen/dzyjZL

Wardormeur: :cry:

PR #13, opened 2017-07-25
[#149099333] Mary wants to not book into past events

grabartley:


src/events/cd-event-list.vue
```diff
@@ -100,6 +101,15 @@
...
         return totalEventCapacity === totalTicketsBooked;
       },
+      isPastEvent(event) {
+        let eventStartTime = moment(event.dates[0].startTime);
+        if (event.type === 'recurring') {
+          eventStartTime = moment(event.dates[event.dates.length - 1].startTime);
+        }
+        const now = moment();
+
+        return now.isAfter(eventStartTime);
```
DanielBrierton: Doesn't take into account timezones. When we parse the event with moment, it'll be converted to the user's local timezone, which may end up being ahead of the current time.

For example:
* User is in Italy (UTC+2 in summer time)
* Current local time is `2017-07-26T16:30:00.000Z`
* `event.dates[0].startTime` is `2017-07-26T16:00:00.000Z` (this should be treated as 16:00 Italian time)
* When parsed, the date is `Wed Jul 26 2017 18:00:00 GMT+0200 (CEST)`
* `moment()` (i.e. now) when parsed is `Wed Jul 26 2017 14:30:00 GMT+0200 (CEST)`
* `now.isAfter(eventStartTime)` will now be false, even though it actually is after the event time

Solution is actually pretty straight forward though. You just have to subtract the `utcOffset` from the parsed `eventStartTime`:
```
eventStartTime.subtract(eventStartTime.utcOffset(), 'm');
```

test/unit/specs/events/cd-event-list.spec.js
```diff
@@ -241,6 +242,83 @@ describe('Event list component', () => {
...
+            {
+              startTime,
+            },
+          ],
+        };
+
+        // ACT & ASSERT
+        expect(vm.isPastEvent(mockEvent)).to.equal(false);
+      });
+    });
```
DanielBrierton: We should have a test that tests timezones... I'll get back to you on that one 😅

DanielBrierton: This should help! https://github.com/plaa/TimeShift-js

PR #14, opened 2017-07-26
[#148854873] Mary wants to be able to only view private events when logged...

grabartley:
... in and joined to the Dojo.


src/events/cd-event-list.vue
```diff
@@ -100,6 +100,12 @@
         });
         return totalEventCapacity === totalTicketsBooked;
       },
+      isEventShown(event) {
+        if (!this.isMember) {
```
DanielBrierton: We have an issue here. `isMember` is based off `usersDojos` which is loaded asynchronously. It's very possible Vue will get to rendering the list before we have `isMember` defined, so private events that should be rendered could be missing. The idea would be if this was a computed property, but obviously there's a problem there since we need the event.

My approach here would be to actually extract each listing (`.cd-event-list__event` and everything inside it) into its own component and bind down the `event`, `dojo`, `currentUser` and `usersDojos`. That way we can have a computed property for `isEventShown` that will automatically react when `usersDojos` is loaded

Like:
```
<event-list-item v=for="event in events" :event="event" :dojo="dojo" :users-dojos="usersDojos" :user="currentUser"></event-list-item>
```

The implementation shouldn't be too big a deal. The main issue will be reorganising the unit tests to split things out.

PR #17, opened 2017-07-26
[#149242487] As Mary I want to be able to return to my search results…

grabartley:
… after I navigate to the Dojo Details page

### DIFF
--- a/src/dojos/cd-find-dojo.vue
+++ b/src/dojos/cd-find-dojo.vue
@@ -39,9 +39,7 @@
       </div>
       <dojo-map :center="coordinates" :class="{ 'cd-find-dojo__results-map': true, 'cd-find-dojo__results-map--hidden': !showMap }" :dojos="allActiveDojos"></dojo-map>
     </div>
-
   </div>
-
 </template>
 <script>
   import Vue from 'vue';
@@ -90,6 +88,7 @@
           this.coordinates.latitude = position.coords.latitude;
           this.coordinates.longitude = position.coords.longitude;
           this.getDojosByLatLong();
+          this.$router.push({ path: '/?currentLocation=true' });
         }, () => {
           this.detectingLocation = false;
         });
@@ -107,6 +106,7 @@
           .then((coords) => {
             this.coordinates = coords;
             this.getDojosByLatLong();
+            this.$router.push({ path: `/?q=${this.searchCriteria}` });
           });
       },
       getAllDojos() {
@@ -137,6 +137,13 @@
         this.coordinates.latitude = this.lat;
         this.coordinates.longitude = this.long;
         this.getDojosByLatLong();
+      }
+      if (this.$route.query.q) {
+        this.searchCriteria = this.$route.query.q;
+        this.searchDojosByAddress();
+      }
+      if (this.$route.query.currentLocation) {
+        this.getCurrentLocation();
       }
       this.getAllDojos();
     },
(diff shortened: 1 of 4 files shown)


src/dojos/cd-find-dojo.vue
```diff
@@ -90,6 +88,7 @@
           this.coordinates.latitude = position.coords.latitude;
           this.coordinates.longitude = position.coords.longitude;
           this.getDojosByLatLong();
+          this.$router.push({ path: '/?currentLocation=true' });
```
DanielBrierton: A cleaner option would be:
```
this.$router.push({ query: { currentLocation: true } });
```

src/dojos/cd-find-dojo.vue
```diff
@@ -107,6 +106,7 @@
           .then((coords) => {
             this.coordinates = coords;
             this.getDojosByLatLong();
+            this.$router.push({ path: `/?q=${this.searchCriteria}` });
```
DanielBrierton: Again, a cleaner option:
```
this.$router.push({ query: { q: this.searchCriteria } });
```

DanielBrierton: Should also probably actually move this up to the top of the `searchDojosByAddress` function. That way, we change the URL immediately.

The one for detect location is fine since it'll wait until the user accepts to share their location before it'll change the URL, but here, we should be able to hit refresh immediately after we search and re-execute the same search.

src/dojos/cd-find-dojo.vue
```diff
@@ -138,6 +138,13 @@
         this.coordinates.longitude = this.long;
         this.getDojosByLatLong();
       }
+      if (this.$route.query.q) {
+        this.searchCriteria = this.$route.query.q;
+        this.searchDojosByAddress();
+      }
+      if (this.$route.query.currentLocation) {
```
DanielBrierton: Better to have `else if` here, because otherwise we could potentially end up loading two different sets of Dojos. I'd be inclined to take `q` over `currentLocation` as it's specific user input.

test/unit/specs/dojos/cd-find-dojo.spec.js
```diff
@@ -166,6 +177,12 @@ describe('The Find dojo vue ', () => {
     vm.long = undefined;
     sandbox.stub(vm, 'getDojosByLatLong');
     sandbox.stub(vm, 'getAllDojos');
+    vm.$route = {
```
DanielBrierton: Should be tests for when `q` and `currentLocation` are defined. 3 combinations to be tested.

test/unit/specs/dojos/cd-find-dojo.spec.js
```diff
@@ -120,12 +120,17 @@ describe('The Find dojo vue ', () => {
...
+    const vm = vueUnitHelper(cdFindDojo());
+    vm.$router = {
+      push: sandbox.spy(),
+    };
     vm.getCurrentLocation();
     requestAnimationFrame(() => {
       expect(vm.detectingLocation).to.equal(true);
       expect(vm.coordinates.latitude).to.equal(10);
       expect(vm.coordinates.longitude).to.equal(89);
+      expect(vm.$router.push).to.have.been.calledOnce;
```
DanielBrierton: Check what it's called with.

test/unit/specs/dojos/cd-find-dojo.spec.js
```diff
@@ -199,6 +219,7 @@ describe('The Find dojo vue ', () => {
         longiutide: 84,
       });
       expect(vm.getDojosByLatLong).to.have.been.calledOnce;
+      expect(vm.$router.push).to.have.been.calledOnce;
```
DanielBrierton: Check what it's called with.

PR #19, opened 2017-07-27
[#149417649] Mary wants to only see next event in series on Dojo Details page

grabartley:


src/events/util.js
```diff
@@ -12,4 +12,15 @@ export default {
     const recurrence = recurrences[event.recurringType];
     return `${i18n.t(recurrence)} ${i18n.t('on')} ${dayName}`;
   },
+  getNextStartTime(event) {
+    const now = moment();
+    const nextDateInfo = event.dates.find((dateInfo) => {
+      const dateMoment = moment(dateInfo.startTime);
+      return dateMoment.isAfter(now);
```
DanielBrierton: I know you just moved this code from one place to another, but it's going to have the same issue as outlined in https://github.com/CoderDojo/cp-zen-frontend/pull/13#discussion_r129610388

wdio.conf.js
```diff
@@ -126,7 +126,7 @@ exports.config = {
     jasmineNodeOpts: {
         //
         // Jasmine default timeout
-        defaultTimeoutInterval: 60000,
+        defaultTimeoutInterval: 120000,
```
DanielBrierton: 😭  We'll get this sorted proper...

PR #24, opened 2017-08-03
[#149671741] As Mary I want to view the frequency, day and time of the Dojo

campbs:


campbs:
Requires CoderDojo/cp-translations#20

PR #29, opened 2017-08-05
fix links overflowing from sidebar on dojo page

butlerx:
screen shot 2017-08-05 at 6 37 34 pm


Wardormeur:
Godammit @butlerx we have stories for that :p https://github.com/CoderDojo/cp-zen-frontend/pull/33

PR #30, opened 2017-08-08
[#149977212] Dojo frequency not appearing in search results and map p…

grabartley:
…opups

### DIFF
--- a/json-server/db.js
+++ b/json-server/db.js
@@ -627,8 +627,8 @@ module.exports = {
       recurringType: 'weekly',
       dates: [
         {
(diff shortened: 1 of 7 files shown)


json-server/db.js
```diff
@@ -627,8 +627,8 @@ module.exports = {
       recurringType: 'weekly',
       dates: [
         {
-          startTime: '2017-08-06T16:30:00.000Z',
```
Wardormeur: I believe your issue is that the creation date goes past after some time : create a new Date instead of changing the month :)

grabartley: Done :)

PR #31, opened 2017-08-08
[#149977380] Twitter and Facebook links are broken on Dojo Details page

grabartley:


src/dojos/cd-dojo-details.vue
```diff
@@ -127,6 +127,20 @@
...
+      buildTwitterLink() {
+        const twitterData = this.dojoDetails.twitter;
+        if (/^[a-zA-Z0-9_@]{1,15}$/.test(twitterData)) {
+          return `https://twitter.com/${twitterData}`;
+        }
+        return twitterData;
+      },
+      buildFacebookLink() {
+        const facebookData = this.dojoDetails.facebook;
+        if (/^[a-zA-Z0-9.]{1,}$/.test(facebookData)) {
```
Wardormeur: Don't you need to escape the "." here?

grabartley: I don't think so, the regex seems to work when tried in the console with values containing '.'

Wardormeur: Could you add a test for it, like you did for the "@" ?

grabartley: Will write a test for this :)

Wardormeur:
Seems like dots in [ ] already escape the dot so that it doesnt represent "any character". However it's still necessary for ( )

PR #32, opened 2017-08-08
[#150044243] Shown address should be the user entered one

grabartley:


src/dojos/cd-dojo-details.vue
```diff
@@ -106,7 +106,17 @@
     },
     computed: {
       address() {
-        return !this.dojoDetails.address1 ? undefined : [
+        if (!this.dojoDetails.address1) {
```
Wardormeur: do you have a scenario where this happen ? It really shouldnt.

src/dojos/cd-dojo-details.vue
```diff
@@ -106,7 +106,17 @@
...
+          return undefined;
+        } else if (this.dojoDetails.address1.includes(this.dojoDetails.countryName)) {
+          return this.dojoDetails.address1;
+        } else if (this.dojoDetails.address1.includes(this.dojoDetails.placeName)) {
```
Wardormeur: placeName is the address returned by GAPI. i wouldn't use it as a reference as it may totally differ

src/dojos/cd-dojo-details.vue
```diff
@@ -106,7 +106,17 @@
...
+          return [
+            this.dojoDetails.address1,
+            this.dojoDetails.countryName,
+          ].join(', ');
+        }
+        return [
```
Wardormeur: My thoughts here are more like : either use the address or the placeName, preferably the address.
I don't think there is need for such extensive checks

src/dojos/cd-dojo-details.vue
grabartley: I don't see one, I just kept it since it was an existing check in the code

src/dojos/cd-dojo-details.vue
grabartley: Okay, I did think the function was too long. I was just trying to account for when address1 doesn't include the place or country because some dojos seem to be that way. I will just use address1 instead :)

PR #33, opened 2017-08-08
[#150044120] Website field on Dojo Details page doesn't wrap properly…

grabartley:
… if above a certain length


json-server/db.js
```diff
@@ -791,8 +791,8 @@ module.exports = {
       recurringType: 'weekly',
       dates: [
         {
-          startTime: '2017-08-06T16:30:00.000Z',
```
Wardormeur: already done in another PR

PR #34, opened 2017-08-09
[#149552987] Mary wants to see last event in series on Dojo Details page

grabartley:
New String here: https://github.com/CoderDojo/cp-translations/pull/22

### DIFF
--- a/src/events/cd-event-list-item.vue
+++ b/src/events/cd-event-list-item.vue
@@ -20,10 +20,6 @@
         <div class="cd-event-list-item__times-timestamp">
           {{ event.dates[0].startTime | cdTimeFormatter }} - {{ event.dates[0].endTime | cdTimeFormatter }}
         </div>
-        <div v-if="isRecurring" class="cd-event-list-item__recurring-info">
-          <span class="fa fa-info-circle cd-event-list-item__recurring-info-icon"></span>
-          <span class="cd-event-list-item__recurring-info-text">{{ `${recurringFrequencyInfo} ${$t('at')}` }} {{ event.dates[0].startTime | cdTimeFormatter }} - {{ event.dates[0].endTime | cdTimeFormatter }}</span>
-        </div>
       </div>
     </div>
     <div v-if="canBook && !isPastEvent" class="cd-event-list-item__view-wrapper">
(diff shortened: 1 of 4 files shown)


src/events/cd-event-list-item.vue
```diff
@@ -36,6 +32,14 @@
         {{ isFull ? $t('Full') : $t('See Details and Book') }}
       </router-link>
     </div>
+    <div v-if="isRecurring" class="cd-event-list-item__recurring-info">
+      <span class="fa fa-info-circle cd-event-list-item__recurring-info-icon"></span>
+      <span class="cd-event-list-item__recurring-info-header">{{ $t('This is a recurring event') }}</span>
+      <div class="cd-event-list-item__recurring-info-text">
+        {{ `${recurringFrequencyInfo} ${$t('at')}` }} {{ event.dates[0].startTime | cdTimeFormatter }} - {{ event.dates[0].endTime | cdTimeFormatter }},
```
Wardormeur: I'm pretty sure the order of dates is not necessarly respected

src/events/cd-event-list-item.vue
```diff
@@ -36,6 +32,14 @@
...
+        {{ $t('from') }} {{ event.dates[0].startTime | cdDateFormatter }} {{ $t('to') }} {{ event.dates[event.dates.length - 1].startTime | cdDateFormatter }}
```
Wardormeur: Translation-wise, this cannot work

src/events/cd-event-list-item.vue
grabartley: So event.dates[0] is not always the earliest date?

Wardormeur: Neither is dates[lenght-1] necessarly the latest.

src/events/cd-event-list-item.vue
grabartley: I only did it this way because filters cannot be used in String interpolation and are used to format the date and time

Wardormeur: Use the filter into a computed property instead :)

src/events/cd-event-list-item.vue
grabartley: Oh, I didn't know that. Doesn't that mean next event in series is broken too since it finds the first date in the array that is after today but that may not necessarily be the soonest then?

src/events/cd-event-list-item.vue
grabartley: Okay I will try this :)

src/events/cd-event-list-item.vue
Wardormeur: It's an edge case that happened at least once. My guess is it'll work 98% of the time, but I've seen in the past events with the latest date not being the latest. Could have been a temp glitch too, but considering we're manipulating an array of dates, i'd wouldn't rely on the index to order them.

grabartley: Okay, I will try to find a different way :)

PR #40, opened 2017-08-15
[#149977212] getDojosByLatLong now returns a camelCase response

grabartley:


src/dojos/service.js
```diff
@@ -24,13 +25,20 @@ const DojosService = {
...
+    const dojos = { body: [] };
+    const response = await Vue.http.post(`${Vue.config.apiServer}/api/2.0/dojos/search-bounding-box`, {
+      query: {
+        lat,
+        lon,
+        radius,
+      },
+    });
+    response.body.forEach((snakeCaseDojo) => {
+      dojos.body.push(mapKeys(snakeCaseDojo, (value, key) => camelCase(key)));
```
Wardormeur: you're losing everything which is not "body".
I'd recommend to clone response.body, delete response.body, and then replacing response.body with the newly formatted.

grabartley: Ah yes, will change this now :)

test/unit/specs/dojos/service.spec.js
```diff
@@ -24,7 +24,6 @@ describe('Dojos Service', () => {
   const expectedResult = { name: 'Cool Dojo' };
 
   const expectedDojos = [{
```
Wardormeur: aren't expectedDojos stil lsupposed to have entity$ ?

grabartley: If they do, the test fails because the processed snake case dojos don't have it and they are expected to equal expectedDojos

PR #54, opened 2017-08-25
[#148741643] Investigate making sure new pages scroll to top, not where scroll is on previous page.

campbs:


src/router/index.js
```diff
@@ -85,4 +85,7 @@ export default new Router({
       },
     },
   ],
+  scrollBehavior() {
+    return { x: 0, y: 0 };
```
Wardormeur: According to documentation, this will also overwrite anchor navigation.
To be added as an improvement.

campbs: Unclear what you mean by this. What has to be added?

DanielBrierton: Should probably follow this piece of doc instead:

> Returning the savedPosition will result in a native-like behavior when navigating with back/forward buttons:
> ```
> scrollBehavior (to, from, savedPosition) {
>   if (savedPosition) {
>     return savedPosition
>   } else {
>     return { x: 0, y: 0 }
>   }
> }
> ```

Wardormeur: My point is that url#myLink won't work with those. We used to use them for search dojo results, or specific section post-save that you want to focus on.
Considering that's also a quite common (and expected) behavior, it's normally nice to have to avoid the "how the hell is this not working?" scenario

DanielBrierton: Then I guess it should be this:
```javascript
scrollBehavior (to, from, savedPosition) {
  if (savedPosition) {
    return savedPosition;
  } else if (to.hash) {
    return {
      selector: to.hash,
    };
  } else {
    return { x: 0, y: 0 };
  }
}
```

DanielBrierton: We can also specify an offset for when going the the hash, which will sort out our fixed header issue

Wardormeur: @campbs @grabartley Add an id to each section of dojo-detail page so we can hotlink the "upcoming event" section for exemple  and test this out with the offset :)

src/router/index.js
```diff
@@ -13,6 +13,16 @@ Vue.use(Router);
 
 export default new Router({
   mode: 'history',
+  scrollBehavior: (to, from, savedPosition) => {
+    if (to.hash) {
```
Wardormeur: I would personally put savedPosition first if it doesn't create issues.
Reason is : you're navigating from a page with a hash, you should prob be scrolling where you were, not at the hash you arrived

campbs: We did it this way because when loading a page with a hash with savedPosition first, it wasn't jumping to the hash since savedPosition was taking precedence

DanielBrierton: savedPosition should take precedence over the hash 😄

Wardormeur:
@campbs @grabartley conflicts !

PR #58, opened 2017-09-04
[#148105343] Fixed issue with totalPages

campbs:


src/dojos/cd-dojo-list.vue
```diff
@@ -55,13 +55,15 @@
...
     watch: {
       dojos() {
-        // if the page number in the route exceeds the number of valid pages
-        if (this.$route.query.p > this.$refs.pagination.totalPages) {
-          this.goToPage(this.$refs.pagination.totalPages);
-        // if the page number in the route is less than one
-        } else if (this.$route.query.p < 1) {
-          this.goToPage(1);
-        }
+        Vue.nextTick(() => {
```
DanielBrierton: I think it would be good to add a comment on why `Vue.nextTick` is needed here. Not part of this PR, but an explanation should also be added to `paginatedDojos` watcher too.

PR #59, opened 2017-09-07
[#150788485] Mary wants to follow her Dojos events

grabartley:
Strings: https://github.com/CoderDojo/cp-translations/pull/30

### DIFF
--- a/json-server/users-profile.js
+++ b/json-server/users-profile.js
@@ -1,0 +1,14 @@
+module.exports = {
+  'parent1@example.com': {
+    dob: '1976-05-22T00:00:00.000Z',
+  },
+  'parent2@example.com': {
+    dob: '1969-11-26T00:00:00.000Z',
+  },
+  'child1o13@example.com':{
+    dob: '2000-10-26T00:00:00.000Z',
+  },
+  'admin@coderdojo.org': {
+    dob: '1996-11-22T00:00:00.000Z',
+  },
+};
--- a/json-server/users-profile.js
+++ b/json-server/users-profile.js
@@ -6,7 +6,7 @@ module.exports = {
     dob: '1969-11-26T00:00:00.000Z',
   },
   'child1o13@example.com':{
-    dob: '2000-10-26T00:00:00.000Z',
+    dob: new Date((new Date().getFullYear() - 15).toString()),
   },
   'admin@coderdojo.org': {
     dob: '1996-11-22T00:00:00.000Z',
--- a/src/events/cd-booking-create-account.vue
+++ b/src/events/cd-booking-create-account.vue
@@ -66,6 +66,7 @@
   import { extend, omit, cloneDeep } from 'lodash';
   import VueRecaptcha from 'vue-recaptcha';
   import UserService from '@/users/service';
+  import UserUtils from '@/users/util';
   import EventsService from '@/events/service';
   import DojoService from '@/dojos/service';
   import StoreService from '@/store/store-service';
@@ -145,8 +146,9 @@
         return UserService.getCurrentUser().then((response) => {
           const user = response.body.user;
           const selectedEvent = StoreService.load('selected-event');
-          // TODO: swap initUserType with user type based off dob
-          return DojoService.joinDojo(user.id, selectedEvent.dojoId, [this.user.initUserType.name]);
+          // NOTE : u13 are not supposed to have an user account and hence cannot join
+          const userType = UserUtils.isYouthOverThirteen(new Date(user.dob)) ? 'attendee-o13' : 'parent-guardian';
+          return DojoService.joinDojo(user.id, selectedEvent.dojoId, [userType]);
         });
       },
       bookTickets() {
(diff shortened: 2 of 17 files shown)


src/events/cd-event-list.vue
```diff
@@ -5,15 +5,19 @@
         <div class="cd-event-list__no-events-header">
           {{ $t('No Upcoming Events') }}
         </div>
-        <p class="cd-event-list__no-events-content" v-html="$t('There are no upcoming events planned for this Dojo. Please email {email} if you have any questions.', { email: '<a href=\'mailto:' + dojo.email + '\'>' + dojo.email + '</a>' })">
-        </p>
+        <p v-if="dojo.private" class="cd-event-list__no-events-content" v-html="$t('There are no upcoming events planned for this Dojo. Please email {email} if you have any questions.', { email: '<a href=\'mailto:' + dojo.email + '\'>' + dojo.email + '</a>' })"></p>
```
Wardormeur: Splitting the string in 2 sentences would allow to not duplicate some part of the DOM and save some translations work I believe.

src/users/service.js
```diff
@@ -14,6 +14,8 @@ const UserService = {
     user,
   }).then(() => UserService.login(user.email, user.password)),
 
+  userProfileData: () => Vue.http.post(`${Vue.config.apiServer}/api/2.0/profiles/user-profile-data`, {}),
```
Wardormeur: pretty sure this endpoint shoudln't be called with empty body but rather with a query

test/e2e/specs/dojo-details.js
```diff
@@ -347,6 +352,40 @@ describe('Dojo details page', () => {
...
+    FindDojoPage.openDojoWithQuery('dublin', 5);
+    DojoDetailsPage.joinButton.waitForVisible();
+    DojoDetailsPage.joinButton.click();
+    expect(browser.alertText()).to.equal('Congratulations, you have now joined the Dojo.');
+    browser.alertAccept();
+
+    browser.deleteCookie();
+  });
+
+  it('should not allow someone to join a private dojo', () => {
```
Wardormeur: @rosa to add tests after definition for private event in public dojos

json-server/users-profile.js
```diff
@@ -0,0 +1,14 @@
+module.exports = {
+  'parent1@example.com': {
+    dob: '1976-05-22T00:00:00.000Z',
+  },
+  'parent2@example.com': {
+    dob: '1969-11-26T00:00:00.000Z',
+  },
+  'child1o13@example.com':{
+    dob: '2000-10-26T00:00:00.000Z',
```
DanielBrierton: Tests using this child will break next year 😅

src/events/cd-booking-create-account.vue
```diff
@@ -145,6 +145,7 @@
         return UserService.getCurrentUser().then((response) => {
           const user = response.body.user;
           const selectedEvent = StoreService.load('selected-event');
+          // TODO: swap initUserType with user type based off dob
```
DanielBrierton: Needs to be done first.

src/events/cd-event-list.vue
```diff
@@ -38,18 +43,38 @@
             this.currentUser = response.body.user;
           });
       },
+      loadUsersProfile() {
```
DanielBrierton: While we're at it... async/await? 😛

src/events/cd-event-list.vue
```diff
@@ -38,18 +43,38 @@
...
     },
     watch: {
       currentUser(newUser) {
         if (newUser) {
-          DojosService.getUsersDojos(newUser.id, this.dojo.id).then((response) => {
-            this.usersDojos = response.body;
-          });
+          const userDojoPromise = this.loadUserDojoRole();
+          const userProfilePromise = this.loadUsersProfile();
+          Promise.all([userDojoPromise, userProfilePromise]);
```
DanielBrierton: What's the purpose of this `Promise.all`? It's not returned and there's no `then`. It's unnecessary.

src/events/cd-event-list.vue
```diff
@@ -84,6 +109,23 @@
         color: #7b8082;
         margin-top: 8px;
       }
+      &-join-button {
+        font-size: @font-size-medium;
+        font-weight: bold;
+        margin-top: 16px;
+        padding: 8px;
+        color: @cd-blue;
```
DanielBrierton: Should probably just use `btn-primary`?

json-server/users-profile.js
Wardormeur: true

src/events/cd-booking-create-account.vue
Wardormeur: how is this related to join ?

src/events/cd-event-list.vue
Wardormeur: fair point

src/events/cd-booking-create-account.vue
DanielBrierton: Because it joins as a specific user type. As the code is here, it'll join based off the init user type, which could be champion or mentor for legacy users. The idea here is to limit it to only parent-guardian/attendee-o13 based of the DoB of the logged in user.

PR #63, opened 2017-09-17
Story/i want to cancel my booking

Wardormeur:
Missing : 
 - e2e

Requires : 
 - events : https://github.com/CoderDojo/cp-events-service/pull/161
 - translations :  https://github.com/CoderDojo/cp-translations/pull/33
 - platform : https://github.com/CoderDojo/cp-zen-platform/pull/1193
 https://github.com/CoderDojo/cp-zen-platform/pull/1185 (closed)


.nvmrc
```diff
@@ -1 +1 @@
-lts/boron
```
butlerx: this file isnt really needed anymore

circle.yml
```diff
@@ -1,6 +1,6 @@
 machine:
   node:
-    version: v6.10.3
+    version: v8.6.0
```
butlerx: dump the dockerfile aswell

src/events/cd-user-ticket-list-item.vue
```diff
@@ -0,0 +1,295 @@
+<template>
```
DanielBrierton: I know you already mentioned this, but there are a lot of parallels between this file and cd-event-list-item. I'd definitely rather we look at how to share as much code as possible. I guess the main common items are the border, header, time and recurring notice.

src/events/cd-user-ticket-list-item.vue
```diff
@@ -0,0 +1,295 @@
...
+        <h4 class="cd-user-ticket-list-item__view-application-title"><i class="fa fa-ticket"></i>{{ $t('Tickets') }}</h4>
+        <div v-for="application in applications">
+          <cd-attendee :application="application" :session="sessions[application.sessionId]" :ticket="tickets[application.ticketId]" :user="users[application.userId]"></cd-attendee>
+        </div>
+        <button tag="button"
+          @click="cancel()"
+          class="btn btn-lg cd-user-ticket-list-item__view-cancel-button">
+          {{ $t('Cancel ticket', applications.length) }}</button>
+      </div>
+      <!-- NOTE : What about awaiting approval tickets ?-->
```
DanielBrierton: Yeah, what about it? :P

src/events/cd-user-ticket-list-item.vue
```diff
@@ -0,0 +1,295 @@
...
+      <div class="cd-user-ticket-list-item__datetime">
+        <div v-if="!isRecurring" class="cd-user-ticket-list-item__date-timestamp">
+          {{ event.dates[0].startTime | cdDateFormatter }}
+        </div>
+        <div v-else>
+          <div class="cd-user-ticket-list-item__date-series">{{ $t('Next in series:') }}</div>
+          <span class="cd-user-ticket-list-item__date-timestamp">{{ nextStartTime |  cdDateFormatter }}</span>
+        </div>
+        <div class="cd-user-ticket-list-item__times-timestamp">
+          {{ event.dates[0].startTime | cdTimeFormatter }} - {{ event.dates[0].endTime | cdTimeFormatter }}
```
DanielBrierton: Might as well use `formattedStartTime` and `formattedEndTime` here instead. IIRC filters are gonna be deprecated, so all the better really.

DanielBrierton: And yes, I am aware that you probably just copy and pasted cd-event-list-item and it had this already, but still 😛

src/events/cd-user-ticket-list-item.vue
```diff
@@ -0,0 +1,295 @@
...
+        const sortedDates = sortBy(this.event.dates, date => date.startTime);
+        return this.$options.filters.cdDateFormatter(sortedDates[0].startTime);
+      },
+      formattedLastDate() {
+        const sortedDates = sortBy(this.event.dates, date => date.startTime);
+        return this.$options.filters.cdDateFormatter(sortedDates[sortedDates.length - 1].startTime);
+      },
+    },
+    methods: {
+      prepareSessions() {
```
DanielBrierton: Should probably just be a `sessions` computed property. Same with `prepareTickets` below.

src/events/cd-user-ticket-list-item.vue
```diff
@@ -0,0 +1,295 @@
...
+        const payload = [];
+        const commonFields = ['dojoId', 'eventId', 'sessionId', 'ticketName',
+          'ticketType', 'ticketId', 'userId', 'notes', 'deleted'];
+        this.applications.forEach((application) => {
+          application.deleted = true; // eslint-disable-line no-param-reassign
+          payload.push(pick(application, commonFields));
+        });
+        await EventService.manageTickets(payload);
+        this.applications = [];
+        alert(this.$t('Applications cancelled')); // eslint-disable-line no-alert
```
DanielBrierton: Maybe 'Your applications have been successfully cancelled.'?

src/events/cd-user-ticket-list-item.vue
DanielBrierton: The border is used even more. Could be just a global style that can be used, or a really dumb component that encapsulates the style.

test/e2e/page-objects/my-tickets.js
```diff
@@ -0,0 +1,66 @@
...
+      return BasePage.open.call(this, '/dashboard/tickets');
+    },
+  },
+  events: {
+    get() {
+      return $$('.cd-user-ticket-list-item.cd-event-ticket-list__event');
+    },
+  },
+  eventName: {
+    value(index) {
```
DanielBrierton: I know this pattern is copied from elsewhere, but I really dislike the passing of index in here. To me it makes more sense to just have a get and return the list, then access the specific index by a normal `[index]` from the test instead. Only case where it makes sense to pass in an index is when the selector itself is selecting by index, rather than grabbing an item at a position from a list.

test/unit/specs/users/cd-tickets.spec.js
```diff
@@ -0,0 +1,141 @@
...
+    });
+  });
+
+  afterEach(() => {
+    sandbox.restore();
+  });
+
+  describe('methods', () => {
+    describe('loadCurrentUser', () => {
+      it('should load the current user', (done) => {
```
DanielBrierton: `async () => {` instead, then just use `await vm.loadCurrentUser();`, and remove `requireAnimationFrame` and `done()`.

Rinse, repeat. :P

test/unit/specs/users/cd-tickets.spec.js
```diff
@@ -0,0 +1,141 @@
...
+          id: 'foo',
+        };
+        const mockUserChildrenTransformed = {foo: {id: 'foo'}, 1:{id: '1'}, 2: {id: '2'}};
+
+        // ACT
+        vm.loadUsersChildren();
+
+        // ASSERT
+        requestAnimationFrame(() => {
+          console.log(vm.users);
```
DanielBrierton: Bloorp

wdio.conf.js
```diff
@@ -11,6 +11,7 @@ exports.config = {
     //
     specs: [
         './test/e2e/specs/**/*.js'
+        // './test/e2e/specs/my-tickets.js'
```
DanielBrierton: Bloop

wdio.conf.js
```diff
@@ -72,7 +73,7 @@ exports.config = {
     //
     // Set a base URL in order to shorten url command calls. If your url parameter starts
     // with "/", then the base url gets prepended.
-    baseUrl: process.env.TARGET_URL_BASE || "http://localhost:8000/v2",
+    baseUrl: process.env.TARGET_URL_BASE || "http://localhost:8080/v2",
```
DanielBrierton: No need for this. `e2e-with-mocks` sets TARGET_URL_BASE as necessary

index.js
```diff
@@ -62,6 +62,16 @@ exports.register = function (server, options, next) {
 
   server.route({
     method: 'GET',
+    path: '/dashboard/tickets',
+    handler :{
+      file: {
+        path: __dirname + '/dist/index.html'
```
DanielBrierton: rebase and align with new plugin code

json-server/server.js
```diff
@@ -93,6 +102,15 @@ server.post('/api/2.0/users/login', (req, res) => {
   res.send();
 });
 
+server.get('/api/2.0/user/event/:id/applications', (req, res) => {
+  if (req.cookies.loggedIn) {
+    console.log(req.cookies.loggedIn, applications);
```
DanielBrierton: Glorp

json-server/server.js
```diff
@@ -79,6 +80,14 @@ server.post('/api/2.0/profiles/youth/create', (req, res) => {
   res.send(child);
 });
 
+server.get('/api/2.0/profiles/children-for-user/:parentId', (req, res) => {
+  const children = users[req.cookies.loggedIn].children;
+  children.child1.user = Object.assign({}, children.child1);
```
DanielBrierton: Don't limit to just child1?

src/events/cd-user-ticket-list-item.vue
```diff
@@ -0,0 +1,295 @@
...
+      prepareTickets() {
+        this.event.sessions.forEach((s) => {
+          s.tickets.forEach((t) => {
+            this.tickets[t.id] = t;
+          }, this);
+        });
+      },
+      async cancel() {
+        const payload = [];
+        const commonFields = ['dojoId', 'eventId', 'sessionId', 'ticketName',
```
DanielBrierton: Probably move out of the component so we're not initialising every time we call cancel?

src/events/cd-user-ticket-list-item.vue
```diff
@@ -0,0 +1,295 @@
...
+    },
+    filters: {
+      cdDateFormatter,
+      cdTimeFormatter,
+      cdUrlFormatter,
+    },
+  };
+</script>
+<style scoped lang="less">
+    @import "../common/variables";
```
DanielBrierton: Let's nitpick! Two too many spaces! :trollface:

src/events/service.js
```diff
@@ -15,7 +15,11 @@ const EventsService = {
   loadSessions(eventId) {
     return Vue.http.get(`${Vue.config.apiServer}/api/2.0/events/${eventId}/sessions`);
   },
-  bookTickets(applications) {
+  // TODO : new REST endpoint /event/:id/applications or /user/tickets or /user/event/:id/tickets
```
DanielBrierton: This is done, right?

src/events/cd-user-ticket-list-item.vue
Wardormeur: Well, we should prob display them differently. But that's another story? :)

src/events/cd-user-ticket-list-item.vue
Wardormeur: Computing a value means we're adding a watcher. If the resource is fetched once, we don't need computed props

src/events/cd-user-ticket-list-item.vue
Wardormeur: fair

src/events/cd-user-ticket-list-item.vue
Wardormeur: Sure

src/events/cd-user-ticket-list-item.vue
Wardormeur: How come it hasn't been caught by the linter?

test/unit/specs/users/cd-tickets.spec.js
Wardormeur: nothing to see here :p

test/e2e/page-objects/my-tickets.js
Wardormeur: So you would rather pass down to EventName its parent path and get the name from there?

src/events/cd-event-list-item.vue
```diff
@@ -124,111 +83,10 @@
     },
   };
 </script>
-<style scoped lang="less">
-    @import "../common/variables";
+<style scope lang="less">
```
DanielBrierton: You had it right the first time, it's `scoped`

src/events/cd-user-ticket-list-item.vue
```diff
@@ -53,72 +53,31 @@
...
   export default {
     name: 'user-ticket-list-item',
+    mixins: [EventTile],
     props: ['event', 'dojo', 'usersDojos', 'users'],
     data() {
       return {
         applications: [],
         sessions: {},
         tickets: {},
+        cancelFields: ['dojoId', 'eventId', 'sessionId', 'ticketName',
```
DanielBrierton: Does this really need to be a data prop? Could just have `const cancelFields = [..]` above the export? Just saves the unnecessary tracking of get/set.

test/unit/specs/users/cd-tickets.spec.js
```diff
@@ -42,17 +42,15 @@ describe('Events/Tickets list component', () => {
...
-        vm.loadCurrentUser();
+        await vm.loadCurrentUser();
 
         // ASSERT
-        requestAnimationFrame(() => {
-          expect(vm.currentUser).to.deep.equal(mockUser);
-          done();
-        });
+        expect(vm.currentUser).to.deep.equal(mockUser);
+        return Promise.resolve();
```
DanielBrierton: No need to `return Promise.resolve();`. async functions implicitly do this.

Wardormeur: You were right btw, dunno what went wrong first time :|

test/e2e/page-objects/my-tickets.js
DanielBrierton: No, I mean it should be just
```
eventNames: {
  get() {
    return $('.cd-user-ticket-list-item.cd-event-ticket-list__event .cd-user-ticket-list-item__name');
  },
},
```

Then, when wanting to get the _n_th eventName, you just go `MyTicketsPage.eventNames[n]`

DanielBrierton: (It's no biggie, and something we could do as a sweeping refactor instead, since this is far from the only offender)

Wardormeur: I know but this does only work for eventName (a "one to one" relationship). If we try for tickets (a "one to n") for example, it becomes a mess because you have to count the number of overall tickets.
That's why i'd think using the event "as a root" reference to the css selector is better

test/unit/specs/users/cd-tickets.spec.js
DanielBrierton: /me does my _I was right_ dance. 😝

PR #65, opened 2017-09-29
#150737997 Add join button unconditionnaly and redirect to login/register

Wardormeur:


src/events/cd-event-list.vue
```diff
@@ -8,11 +8,18 @@
         <div>
           <span class="cd-event-list__no-events-content" v-html="$t('There are no upcoming events planned for this Dojo.')"></span>
           <span class="cd-event-list__no-events-content" v-html="$t(`${dojo.private ? 'Please email {email} if you have any questions.' : 'Please join this Dojo for updates or email {email}' }`, { email: '<a href=\'mailto:' + dojo.email + '\'>' + dojo.email + '</a>' })"></span>
-          <button @click="joinTheDojo()" v-if="currentUser && !dojo.private && usersDojos.length === 0" class="cd-event-list__no-events-join-button">{{ $t('Join Dojo') }}</button>
         </div>
+        <button @click="joinTheDojo()" v-if="!dojo.private && !isDojoMember" class="cd-event-list__no-events-join-button">{{ $t('Join Dojo') }}</button>
```
butlerx: is there a reason its not in the div anymore?

src/events/cd-event-list.vue
```diff
@@ -54,12 +66,16 @@
...
-        this.loadUserDojoRole();
+        if (this.currentUser) {
+          const userType = UsersUtil.isYouthOverThirteen(new Date(this.usersProfile.dob)) ? 'attendee-o13' : 'parent-guardian';
+          await DojosService.joinDojo(this.currentUser.id, this.dojo.id, [userType]);
+          /* eslint-disable no-alert */
+          alert('Congratulations, you have now joined the Dojo.');
+          /* eslint-enable no-alert */
+          this.loadUserDojoRole();
+        } else {
+          location.href = `/register/account?referer=${this.$route.path}`;
```
butlerx: why register over login?

src/events/cd-event-list.vue
Wardormeur: Yep : block element on new line so that the button isn't "chaining" with the sentence on responsive

src/events/cd-event-list.vue
Wardormeur: Because this page is the one displaying both register and login

src/events/cd-event-list.vue
```diff
@@ -8,11 +8,18 @@
...
         </div>
+        <button @click="joinTheDojo()" v-if="!dojo.private && !isDojoMember" class="cd-event-list__no-events-join-button">{{ $t('Join Dojo') }}</button>
+      </div>
+    </div>
+    <div v-else>
+      <event-list-item v-for="event in events" :key="event.id" :event="event" :dojo="dojo" :users-dojos="usersDojos" :user="currentUser" class="cd-event-list__event"></event-list-item>
+      <div v-if="!dojo.private && !isDojoMember" class="cd-event-list__event-join">
+        <p class="cd-event-list__event-join-description">{{ $t('or') }}</p>
+        <button @click="joinTheDojo()" class="cd-event-list__event-join-button">{{ $t('Join the Dojo') }}</button>
+        <p class="cd-event-list__event-join-description">{{ $t('to get notified of new events') }}</p>
```
DanielBrierton: Hmmm... Concerned how well this works in terms of i18n. The action wouldn't necessarily always be first in other languages. Not really sure what the alternative is though given we need to bind an event to the button...

I guess you could make the string be `{joinDojoStart}Join the Dojo{joinDojoEnd} to get notified of new events` and then use `cd-translation-component-generator` so you can still bind events as kinda messy as it is... That way a translation could be in the form `To get notified of new events {joinDojoStart}Join the Dojo{joinDojoEnd}` without any issues. Not a big fan of `joinDojoStart` and `joinDojoEnd`, but you get the idea.

src/events/cd-event-list.vue
```diff
@@ -78,6 +94,25 @@
 </script>
 <style scoped lang="less">
   @import "../common/variables";
+  .join-button {
```
DanielBrierton: We don't want the mixin to create a `join-button` class, so use `.join-button() {` instead. Parenthesis define it as a mixin only.

PR #70, opened 2017-10-10
#151684306 - Remove loading of events from unverified dojos

Wardormeur:
Story : #151684306 https://www.pivotaltracker.com/story/show/151684306
Related to https://github.com/CoderDojo/cp-zen-platform/pull/1195


test/unit/specs/events/cd-event-list.spec.js
```diff
@@ -259,13 +259,28 @@ describe('Event list component', () => {
...
+        const vm = vueUnitHelper(eventList());
+        sandbox.stub(vm, 'loadEvents');
+        sandbox.stub(vm, 'loadCurrentUser');
+        vm.dojo = { verified: 0 };
+        MockUsersService.getCurrentUser.returns({ body: { user: { id: '34174952-8ca4-4189-b8cb-d383e3fde992' } } });
+        // ACT
+        vm.$lifecycleMethods.created();
+
+        // ASSERT
+        expect(vm.loadEvents).not.to.have.been.calledOnce;
```
DanielBrierton: This will check if `vm.loadEvents.callCount !== 1`, which means if it's called twice (or more) for some reason, the test will still pass. Should just be 

```javascript
expect(vm.loadEvents).not.to.have.been.called;
```

Same with `vm.loadCurrentUser` below.

Wardormeur: good catch

PR #83, opened 2017-11-14
Missing email field in Create a CoderDojo Account section

Wardormeur:
Solves https://trello.com/c/htQm2Bn4

### DIFF
--- a/src/events/cd-booking-create-account.vue
+++ b/src/events/cd-booking-create-account.vue
@@ -10,6 +10,14 @@
     </div>
     <div class="cd-booking-create-account__container">
       <div class="row">
+        <label class="cd-booking-create-account__label" for="email">{{ $t('Email Address') }}</label>
+        <input type="email" :placeholder="$t('Email address')" class="form-control"
+          name="email" id="email" data-vv-as="email" v-validate="'required|email'" v-model="email">
+        <label class="cd-booking-create-account__email-error text-danger" v-show="errors.has('email:required')">{{ $t('Parent email address is required') }}</label>
+        <label class="cd-booking-create-account__email-error text-danger" v-show="errors.has('email:email')">{{ $t('Parent email address is invalid') }}</label>
+
+      </div>
+      <div class="row">
         <label class="cd-booking-create-account__label" for="password">{{ $t('Password') }}</label>
         <input type="password" class="form-control" placeholder="Password" name="password" id="password" data-vv-as="password"
                v-validate="'required|confirmed:confirmPassword|cd-password'" v-model="password"/>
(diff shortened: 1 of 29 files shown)


src/events/cd-booking-create-account.vue
```diff
@@ -10,6 +10,14 @@
     </div>
     <div class="cd-booking-create-account__container">
       <div class="row">
+        <label class="cd-booking-create-account__label" for="email">{{ $t('Email Address') }}</label>
+        <input type="email" :placeholder="$t('Email address')" class="form-control"
+          name="email" id="email" data-vv-as="email" v-validate="'required|email'" v-model="email">
+        <label class="cd-booking-create-account__email-error text-danger" v-show="errors.has('email:required')">{{ $t('Parent email address is required') }}</label>
```
DanielBrierton: This could be an O13's email, not specifically parent

src/events/cd-booking-create-account.vue
```diff
@@ -118,13 +128,19 @@
...
           .then(this.addChildren)
           .then(this.joinDojo)
           .then(this.bookTickets);
       },
-      async register() {
+      formatProfile() {
         this.profile = StoreService.load(`booking-${this.eventId}-user`);
+        this.profile.email = this.email;
+        StoreService.save(`booking-${this.eventId}-user`, this.profile);
+        return Promise.resolve();
```
DanielBrierton: Why? There's nothing async here. submitAccount should just be
```javascript
async submitAccount() {
  this.formatProfile();
  return this.register()
    .then(this.addChildren)
    .then(this.joinDojo)
    .then(this.bookTickets);
}
```

If you _really_ want to chain them, then just make this function async and the returned Promise is implicit.

src/events/cd-booking-parent-form.vue
```diff
@@ -30,11 +30,6 @@
     <p class="cd-booking-parent-form__phone-error text-danger" v-show="errors.has('phoneNumber:required')">{{ $t('Phone number is required') }}</p>
     <p class="cd-booking-parent-form__phone-error text-danger" v-show="errors.has('phoneNumber:numeric')">{{ $t('Phone number is invalid') }}</p>
 
-    <label class="cd-booking-parent-form__label" for="email">{{ $t('Email Address') }}</label>
```
DanielBrierton: This field isn't supposed to be removed. The consensus in Pivotal, and Conor has since agreed (although I do concede that the story isn't updated) is that the field in the account section should sync with this one and be readonly. That way it's clear who the account is being created for. We could leave them separate for now and have another story to sync them up, but at the very least, this field should be still here.

DanielBrierton:
Closing in favour of #100

PR #91, opened 2018-01-11
Add new languages

Wardormeur:


src/locale/cd-lang-picker.vue
```diff
@@ -132,11 +55,13 @@
...
     },
-    created() {
+    async created() {
+      this.availableLanguages = (await this.getAvailableLanguages()).body;
       const browserLocale = navigator.language.replace('-', '_');
       const matchingLocale = find(this.availableLanguages,
-        language => language.value === browserLocale);
+        language => language.code === browserLocale);
       const langCookie = Cookie.get('NG_TRANSLATE_LANG_KEY');
+      console.log('debug', langCookie, matchingLocale, browserLocale);
```
DanielBrierton: *cough*

Wardormeur: *nothing to see here*

PR #98, opened 2018-03-05
Add display of past events if no future event is available

Wardormeur:
Show up the past events when the dojo has no future events.

## Description
I tried to keep the active list independent but the message, implemented in the cd-event-list as well, changes depending on weither the dojo has active events or not. That makes the (active)event-list dependent on the (past)event-list without explicit visbility.
A solution would be to separate the loading of events into the container and have the state into the container rather. A dumb cd-event-list could then be used to take as params the info and loop over it ?

## Motivation and Context

https://trello.com/c/zRuyrmEQ

## How Has This Been Tested?

Mock server

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.
- [x] I gave a thought about security (XSS, perms, injection of params, scoping, ...)
- [x] Linting passes

## Related PRs

TODO: cp-translations

## Reviewer checklist:

 - [ ] Security
   - [ ] Permissions
   - [ ] Data scope and leaks
   - [ ] Injection
   - [ ] XSS
 - [ ] Performance
   - [ ] tested with production data
 - [ ] Code
   - [ ] Single responsibility
   - [ ] Naming
 - [ ] Design
   - [ ] Responsive
   - [ ] Scale well with production data
   - [ ] Looks great !


Wardormeur:
TODO : moment().unix() instead of .format('X')

src/events/cd-event-list.vue
```diff
@@ -88,7 +131,10 @@
     },
     created() {
       if (this.dojo.verified) {
-        this.loadEvents();
+        this.loadEvents()
```
DanielBrierton: `this.futureEvents = (await this.loadEvents()).body.results;`?

src/events/cd-event-list.vue
```diff
@@ -1,52 +1,78 @@
...
     </div>
-    <div v-else>
-      <event-list-item v-for="event in events" :key="event.id" :event="event" :dojo="dojo" :users-dojos="usersDojos" :user="currentUser" class="cd-event-list__event"></event-list-item>
-      <div v-if="!dojo.private && !isDojoMember" class="cd-event-list__event-join">
-        <p class="cd-event-list__event-join-description">{{ $t('or') }}</p>
-        <button v-ga-track-click="'join_dojo_has_events'" @click="joinTheDojo()" class="cd-event-list__event-join-button">{{ $t('Join the Dojo') }}</button>
-        <p class="cd-event-list__event-join-description">{{ $t('to get notified of new events') }}</p>
+    <div class="cd-event-list__events">
+      <event-list-item v-for="event in futureEvents" :key="event.id" :event="event" :dojo="dojo" :users-dojos="usersDojos" :user="currentUser" class="cd-event-list__event"></event-list-item>
+      <div v-if="!hasFutureEvents">
```
DanielBrierton: Move hasPastEvents check into here?

src/events/cd-event-list.vue
Wardormeur: No. We don't want the calls to be blocking

PR #100, opened 2018-03-20
Missing email field in Create a CoderDojo Account section

DanielBrierton:
Opening in place of #83


src/events/cd-booking-create-account.vue
```diff
@@ -86,6 +92,9 @@
     data() {
       return {
         profile: null,
+        get email() {
```
DanielBrierton: Should be a computed property instead

DanielBrierton:
Closing due to changes how we're going to handle new accounts

PR #86, opened 2017-11-17
Add country to user registration based upon dojo's for FTBooking

Wardormeur:
https://trello.com/c/V4cnTXw1
Depends on : 
https://github.com/CoderDojo/cp-zen-platform/pull/1222
https://github.com/CoderDojo/cp-users-service/pull/280

PR #101, opened 2018-03-20
Add country to user registration based upon dojo's for FTBooking

DanielBrierton:
Opening in place of #86


DanielBrierton:
Closing due to changes how we're going to handle new accounts

PR #106, opened 2018-03-22
Add 'hosted by'

Wardormeur:
Fix headers styling
Fix "next" button styling

### DIFF
--- a/src/common/styles/cd-primary-button.less
+++ b/src/common/styles/cd-primary-button.less
@@ -1,0 +1,8 @@
+.primary-button() {
+  margin-top: 48px;
+  width: 200px;
+  height: 50px;
+  border-radius: 4px;
+  font-size: 16px;
+  border: none;
+}
--- a/src/common/styles/cd-primary-button.less
+++ b/src/common/styles/cd-primary-button.less
@@ -1,6 +1,7 @@
 .primary-button() {
   margin-top: 48px;
-  width: 200px;
+  min-width: 200px;
+  padding: 0px 16px;
   height: 50px;
   border-radius: 4px;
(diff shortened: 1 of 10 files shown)


Wardormeur:
Depends on https://github.com/CoderDojo/cp-translations/pull/55

src/common/styles/cd-primary-button.less
```diff
@@ -0,0 +1,8 @@
+.primary-button() {
+  margin-top: 48px;
+  width: 200px;
```
DanielBrierton: Becomes a problem when the text is long. I'd propose a min-width (with padding to ensure text doesn't go right to the edge)

DanielBrierton: I realise now that this is copy/paste from elsewhere, but... 😅

src/events/cd-event-details.vue
```diff
@@ -127,6 +152,12 @@
       flex: 4;
       max-width: 340px;
 
+      &--dojo {
+        margin: 24px 0;
+      }
+      &--time {
+        margin: 24px 0 48px 0;
```
DanielBrierton: Why the difference in margin between this and `&--dojo`?

Wardormeur: Because 48px is huge. Also because it is not the same kind of element. "hosted-by" is a pre- of all left-column items with a border-bottom. See ticket screenshot for more :) https://trello.com/c/wMuZBelr

src/common/styles/cd-primary-button.less
Wardormeur: Fair still

PR #109, opened 2018-04-02
Enhancement/manage applications

Wardormeur:
This is a rework of the application management screen.
It still has some issue, but this is a WIP : any comment is welcome.
The next steps will be to simplify the event-applications component by externalizing anything relative to the state as well as externalizing the slots


Wardormeur:
Requires : https://github.com/CoderDojo/events-service/pull/11

PR #108, opened 2018-03-27
Bugfix/returning user booking

Wardormeur:
Move some stuff around
Create a new page for register or login

PR #110, opened 2018-04-03
Set up frontend to run on docker with selenium. Also, set up the new login page as well as the e2e and unit tests for it.

heyitsmaimai:


src/users/cd-login.vue
```diff
@@ -17,13 +36,52 @@
       return {
         email: '',
         password: '',
+        loginFailed: false,
       };
     },
     methods: {
+      async validateForm() {
+        try {
+          return await this.$validator.validateAll();
```
Wardormeur: I believe the try-catch is unecessary here : 
http://vee-validate.logaretm.com/examples.html#validate-form

src/users/cd-login.vue
```diff
@@ -17,13 +36,52 @@
...
       async login() {
-        await UserService.login(this.email, this.password);
-        this.$router.push('/');
+        const valid = await this.validateForm();
+        if (valid) {
+          const response = await UserService.login(this.email, this.password);
+          if (response.body.ok === false) {
+            this.loginFailed = true;
+          } else {
+            this.$router.push(this.$route.query.referer || '/');
```
Wardormeur: In term of cross-compat, this need to be a refresh, not a router.push as it needs to work with the angular app. That's why you're using a normal <a href> for the "reset" link rather than a <router-link> : it needs to reload the page for angular to load.
Also, i'm pretty sure we have 3 different keyword for referer.. "next" is another one. Now, up to us (cc @DanielBrierton ) to decide if we harmonize it.

src/users/cd-login.vue
Wardormeur: https://vee-validate.logaretm.com/api.html#validator-validate Docs says it can throw an error but it's not validation related. At that point, i'd prefer the error to bubble up

src/users/cd-login.vue
```diff
@@ -17,13 +36,52 @@
...
+        }
+      },
       async login() {
-        await UserService.login(this.email, this.password);
-        this.$router.push('/');
+        const valid = await this.validateForm();
+        if (valid) {
+          const response = await UserService.login(this.email, this.password);
+          if (response.body.ok === false) {
+            this.loginFailed = true;
```
Wardormeur: You can use vee-validate's Errorbag to do backend validation https://github.com/CoderDojo/cp-zen-frontend/pull/108/files#diff-c44cabe05813041a88f869b01b78e0c5R177

src/users/cd-login.vue
DanielBrierton: Fair enough. We hit validator throwing an error before, but never really dug into why. It only occurred when something was invalid, so for us 1+1=2, but it obviously seems we were wrong. 😅

src/users/cd-login.vue
DanielBrierton: No, this should be fine. If the route doesn't exist in the Vue app, it will then refresh and defer to the Angular app. If it's in the Vue app, cool, we'll continue.

As for referrer. For starters, I've realised we misspelt refe**rr**er... 😅 Yeah, I wasn't sure if we wanted to settle on just one thing, but obviously, then we'd need to update anywhere in the Angular app where `next` is used, which is a pain. Perhaps we should just support both? Not sure we have anything beyond those two but can check.

test/e2e/specs/login.js
```diff
@@ -0,0 +1,71 @@
...
+		expect(LoginPage.password.isVisible()).to.equal(true);
+		expect(LoginPage.login.isVisible()).to.equal(true);
+		expect(LoginPage.forgotPassword.isVisible()).to.equal(true);
+		expect(LoginPage.forgotPassword.getAttribute('href')).to.equal('/reset');
+		expect(LoginPage.register.isVisible()).to.equal(true);
+		expect(LoginPage.register.getAttribute('href')).to.equal('/register');		
+	});
+
+	it('should show an error when no email is provided on a blur event', () => {
+		LoginPage.email.hasFocus() = true;
```
DanielBrierton: `hasFocus` will return a boolean, so won't be assignable. There's no `setFocus` or `focus` method in webdriver.io, so you should just use `click`, then to blur away from it just click the password field or something.

Come to think of it, I don't think I sent on the docs for webdriver.io. Here's the docs for the click function: http://webdriver.io/api/action/click.html

test/e2e/specs/login.js
```diff
@@ -0,0 +1,71 @@
...
+		expect(LoginPage.email.getText()).to.equal('');
+		LoginPage.email.hasFocus() = false;
+		expect(LoginPage.emailReqError.isVisible()).to.equal(true);
```
DanielBrierton: Might be good to check that it's not visible before we even focus the email field. This doesn't test that it appears after, just that it shows at some point.

test/e2e/specs/login.js
DanielBrierton: This needs to be done for all the below tests where you do the same thing.

test/e2e/specs/login.js
```diff
@@ -0,0 +1,71 @@
...
+
+	it('should show both requirement errors when a login is attempted if no email and no password is provided', () => {
+		LoginPage.login.click();
+		expect(LoginPage.emailReqError.isVisible()).to.equal(true);
+		expect(LoginPage.passwordReqError.isVisible()).to.equal(true);
+	});
+
+	it('should show the password requirement and email format error when a login is attempted if no password and incorrect form of email is provided', () => {
+		LoginPage.email.setValue('janedoe');
+		expect(LoginPage.emailFormatError.isVisible()).to.equal(true);
```
DanielBrierton: You need to click submit before checking for these errors

test/e2e/page-objects/login.js
```diff
@@ -17,9 +27,39 @@ const LoginPage = Object.create(BasePage, {
   },
   login: {
     get() {
-      return $('input[type=submit]');
+      return $('isnput[type=submit]');
```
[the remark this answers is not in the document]
heyitsmaimai: haha yes i noticed and changed that back!

test/e2e/specs/login.js
heyitsmaimai: makes sense, thanks

src/users/cd-login.vue
```diff
@@ -41,7 +41,11 @@
     },
     methods: {
       async validateForm() {
-        return this.$validator.validateAll();
+        try {
```
DanielBrierton: I get why you put this back in, because one of the tests failed. What should actually be done is the test itself should be removed along with this try/catch as it's no longer valid.

src/users/cd-login.vue
```diff
@@ -1,10 +1,29 @@
...
-    </form>
+  <div class="cd-login container-fluid">
+    <div class="row">
+      <h3 class="cd-login__header text-center">Login</h3>
+    </div>
+    <div class="row">
+      <div class="cd-login__box">
+        <form @submit.prevent="login">
+          <div class="form-group">
+            <label>Email</label>
```
DanielBrierton: As you already mentioned, but just adding here as a reminder to both of us, all strings should be wrapped in `{{ $t(...) }}`. They also need to be added to https://github.com/CoderDojo/cp-translations if they haven't already, which I realise now I probably haven't told you... 😅

test/e2e/page-objects/login.js
```diff
@@ -20,11 +30,56 @@ const LoginPage = Object.create(BasePage, {
...
+    get() {
+      return $('.cd-login__login-failed');
+    },
+  },
   open: {
     value() {
       return BasePage.open.call(this, '/v2/login');
     },
   },
+  openWithReferrer: {
```
DanielBrierton: A couple things about this:

1. It won't open the login page with a referrer query param, but will open the referrer directly. You need BasePage.open to be called with `'/v2/login?referrer=%2Fdojos%2Fie%2Fsome%2Fslug'`, but you don't want `'/v2/login?referrer='` when there's no referrer given.
2. I'd actually be inclined to add this to BasePage.open as a generic handler for query parameters. Then open in this file can just take in a query object and pass it down to BasePage.open. If this doesn't make sense to you, grab me and I'll run through it with you.

test/e2e/specs/login.js
```diff
@@ -0,0 +1,116 @@
...
+  it('should login successfully with a referrer query param', () => {
+    LoginPage.open();
+    LoginPage.email.setValue('parent1@example.com');
+    LoginPage.password.setValue('testparent1');
+    LoginPage.login.click();
+    LoginPage.openWithReferrer({ referrer: '/dashboard/tickets' });
+    expect(browser.getUrl()).to.equal('http://localhost:8080/dashboard/tickets');
+  });
+
+  it('should should login successfully without a referrer query param', () => {
```
DanielBrierton: Being super pedantic here, but `should should` 😛

test/e2e/specs/login.js
```diff
@@ -0,0 +1,116 @@
...
+    LoginPage.open();
```
DanielBrierton: Once you've fixed this to open with the referrer param correctly, then you should be opening with the referrer here. The idea is when you click login, you should be brought to the referrer URL. Your tests are passing right now because you're navigating to the referrer URL directly after login. Also, best to test with something other than '/'. '/dojos/ie/smithfield/smithfield-awesome-dojo' should work fine. You can even confirm that the Dojo details page loads afterwards.

test/e2e/specs/login.js
```diff
@@ -0,0 +1,116 @@
...
+    expect(LoginPage.emailReqError.isVisible()).to.equal(false);
+    LoginPage.password.setValue('test123');
+    LoginPage.login.click();
+    LoginPage.emailReqError.waitForVisible();
+    expect(LoginPage.emailReqError.isVisible()).to.equal(true);
+  });
+
+  it('should login successfully with a referrer query param', () => {
+    LoginPage.open();
```
DanielBrierton: Once you've fixed this to open with the referrer param correctly, then you should be opening with the referrer here. The idea is when you click login, you should be brought to the referrer URL. Your tests are passing right now because you're navigating to the referrer URL directly after login.

test/e2e/specs/login.js
```diff
@@ -0,0 +1,116 @@
...
+    LoginPage.openWithReferrer({ referrer: '/dashboard/tickets' });
+    expect(browser.getUrl()).to.equal('http://localhost:8080/dashboard/tickets');
+  });
+
+  it('should should login successfully without a referrer query param', () => {
+    LoginPage.open();
+    LoginPage.email.setValue('parent1@example.com');
+    LoginPage.password.setValue('testparent1');
+    LoginPage.login.click();
+    LoginPage.openWithReferrer({ referrer: '/' });
```
DanielBrierton: This is supposed to test when there's no referrer, but here there is a referrer of '/'.

src/users/cd-login.vue
```diff
@@ -1,10 +1,29 @@
...
+          <div class="form-group">
+            <label>Password</label>
+            <input class="form-control" data-vv-name="password" data-vv-validate-on="blur" type="password" v-model="password" v-validate="'required'" />
+          </div>
+           <p class="cd-login__password-req-err text-danger" v-show="errors.has('password:required')">{{ $t('Password is required') }}</p>
+          <input class="cd-login__button btn btn-primary" type="submit" value="Login" />
+        </form>
+        <p v-show="errors.has('loginFailed')" class="cd-login__login-failed text-danger">{{ $t('There was a problem logging in: Invalid email or password') }}</p>
+        <p class="cd-login__forgot-password"><a href="/reset">{{ $t('Forgot password?') }}</a></p>
+        <p class="cd-login__register">{{ $t("Don't have an account?") }} <a href="/register">{{ $t('Register Here') }}</a></p>
```
Wardormeur: Change "Register Here" to "Register here" for translator's sake

src/users/cd-login.vue
```diff
@@ -1,10 +1,29 @@
...
+          <p class="cd-login__email-req-err text-danger" v-show="errors.has('email:required')">{{ $t('Email is required') }}</p>
+          <p class="cd-login__email-format-err text-danger" v-show="errors.has('email:email')">{{ $t('Email should be in the format: janedoe@example.com') }}</p>      
+          <div class="form-group">
+            <label>Password</label>
+            <input class="form-control" data-vv-name="password" data-vv-validate-on="blur" type="password" v-model="password" v-validate="'required'" />
+          </div>
+           <p class="cd-login__password-req-err text-danger" v-show="errors.has('password:required')">{{ $t('Password is required') }}</p>
+          <input class="cd-login__button btn btn-primary" type="submit" value="Login" />
+        </form>
+        <p v-show="errors.has('loginFailed')" class="cd-login__login-failed text-danger">{{ $t('There was a problem logging in: Invalid email or password') }}</p>
```
Wardormeur: Invalid email or password is an existing string, i'd prefer we use it as a variable so that we can keep the translation live and save the translators some work

Wardormeur: cc @DanielBrierton

Wardormeur: no, more like this -> https://github.com/CoderDojo/cp-zen-frontend/blob/master/src/dojos/cd-dojo-details.vue#L42

heyitsmaimai: I have this {{ $t('There was a problem logging in: {msg}', {msg: 'Invalid email or password!'}) }}. Am I meant to have it as a variable with a different syntax?

PR #111, opened 2018-04-09
Create add child ticket element interface. Add in dependencies for vue multiselect and update dependency for vue-dob-picker. Also commented out mad stuff from event sessions

heyitsmaimai:


src/events/cd-event-add-child-ticket.vue
```diff
@@ -0,0 +1,182 @@
+<template>
+  <div class="cd-child-ticket__ticket-box">
+    <div class="cd-child-ticket__header">
+       <h3 class="cd-child-ticket__ticket-name">{{ $t('Ticket')}} {{firstName|addPosession}} </h3>
```
Wardormeur: Posession?

src/events/cd-event-add-child-ticket.vue
```diff
@@ -0,0 +1,182 @@
...
+    </div>
+    <div class="cd-child-ticket__body">
+      <label>{{ $t('Name')}}</label>
+      <div class="cd-child-ticket__child-name">
+        <input class="cd-child-ticket__first-name form-control" v-model="firstName" type="firstName" :placeholder="$t('First Name')" data-vv-name="firstName" data-vv-validate-on="blur" v-validate="'required'"/>
+        <input class="cd-child-ticket__surname form-control" v-model="surename" type="surname" :placeholder="$t('Surname')" data-vv-name="surname" data-vv-validate-on="blur" v-validate="'required'"/>
```
Wardormeur: surname

src/events/cd-event-add-child-ticket.vue
```diff
@@ -0,0 +1,182 @@
...
+      Multiselect,
+    },
+    data() {
+      return {
+        firstName: '',
+        surame: '',
+        dob: null,
+        genderSelect: '',
+        genderInput: '',
+        whyGender: false,
```
Wardormeur: genderExplaination could be better imho :)

src/events/cd-event-add-child-ticket.vue
```diff
@@ -0,0 +1,182 @@
...
+        ticketTouch: false,
+        tickets: '',
+      };
+    },
+    methods: {
+      showWhy() {
+        this.whyGender = true;
+      },
+      onTicketTouch() {
+        this.ticketTouch = true;
```
Wardormeur: Could make use of http://vee-validate.logaretm.com/validation.html#flags instead

src/events/cd-event-add-child-ticket.vue
```diff
@@ -0,0 +1,182 @@
...
+      },
+      name() {
+        return `${this.firstName} ${this.surame}`;
+      },
+      specifyGender() {
+        return this.genderSelect === 'specify';
+      },
+      gender() {
+        this.whyGender = false;
+        if (this.genderSelect === 'specify') {
```
Wardormeur: You could use the same v-model for both scenario. In the end, i believe it'll just store a string

src/events/cd-event-add-child-ticket.vue
heyitsmaimai: Maybe i used to wrong wording. The filter is meant to ass the child's name onto the ticket. So it would say 'Ticket - Jane' or whatever their name is.

src/events/cd-event-add-child-ticket.vue
heyitsmaimai: I did try that earlier. The flags dont change on touch/touched for vue multiselect it seems? I just used the touch method that uses @close that they had in their docs

src/events/cd-event-add-child-ticket.vue
heyitsmaimai: Tried that! Came across a scenario where if you blurred out of the input field for specifying gender, it would give the required error, and then if you selected another gender (ex male) the error would still show because that required error is still there for the input field even though it wasn't being shown any longer. So this is what was used instead

src/events/cd-event-add-child-ticket.vue
Wardormeur: Then it's possession I believe :)

src/events/cd-event-add-child-ticket.vue
Wardormeur: This one ? https://vue-multiselect.js.org/#sub-custom-configuration

src/events/cd-event-add-child-ticket.vue
```diff
@@ -0,0 +1,182 @@
...
+          return ` - ${firstName}`;
+        }
+        return firstName;
+      },
+    },
+    computed: {
+      childTickets() {
+        return this.sessions.map(session => ({
+          description: session.description,
+          entity$: session.entity$,
```
Wardormeur: You can drop the entity$ from there. It shouldn't have been there in the first place and won't be.

test/unit/specs/events/cd-event-add-child-tickets.spec.js
```diff
@@ -0,0 +1,311 @@
...
+        expect(vm.gender).to.equal('male');
+      });
+    });
+
+    describe('computed.invalidTicket', () => {
+      it('should return the true if the ticketTouch is true and if the ticket length is equal to 0', async () => {
+        // ARRANGE
+        const vm = vueUnitHelper(ChildTicketWithMocks);
+        vm.ticketTouch = true;
+        vm.tickets = '';
```
Wardormeur: This works, but shouldn't it be an array rather than a string ?

src/events/cd-event-add-child-ticket.vue
```diff
@@ -0,0 +1,182 @@
...
+    data() {
+      return {
+        firstName: '',
+        surname: '',
+        dob: null,
+        genderSelect: '',
+        genderInput: '',
+        whyGender: false,
+        ticketTouch: false,
+        tickets: '',
```
Wardormeur: Shouldnt' that be an array rather?

src/events/cd-event-add-child-ticket.vue
```diff
@@ -0,0 +1,182 @@
...
+          <option value="" selected data-default disabled>{{ $t('Select an option') }}</option>
+          <option value="male">{{ $t('Male') }}</option>
+          <option value="female">{{ $t('Female') }}</option>
+          <option value="prefer not to answer">{{ $t('Prefer not to answer') }}</option>
+          <option value="specify">{{ $t('Specify Identity') }}</option>
+        </select>
+        <input v-show="specifyGender" class="form-control" v-model="genderInput" :placeholder="$t('Identify as...')" data-vv-name="gender" data-vv-validate-on="blur" v-validate="specifyGender ? 'required' : ''"/>
+      </div>
+      <p class="cd-child-ticket__gender-err text-danger" v-show="errors.has('gender:required')">{{ $t('Gender is required') }} <a v-on:click="showWhy">{{ $t('Why is this required? Click Here To Find Out More') }}</a><p v-show="whyGender && errors.has('gender:required')">{{ $t(`We want to provide activities that appeal to people regardless of their gender. To check how well we are succeeding, we'd like to find out whether or not people of different genders are equally likely to take part.`) }}</p></p>
```
Wardormeur: Jump line here please :p This is way too long and I'm surprised we don't have a linter crying over that length

Wardormeur: Also a <p> inside a <p> is invalid, use a span inside rather

src/events/cd-event-add-child-ticket.vue
heyitsmaimai: Okeedokes!

test/unit/specs/events/cd-event-add-child-tickets.spec.js
heyitsmaimai: It should, that's my bad. Will fix that

DanielBrierton:
@Wardormeur Yeah, separate component for gender sounds good actually. Exactly what I've done with state in coolest-platform.

```html
<template>
  <div class="row row-no-margin">
    <div class="col-2fr">
      <select class="full-width-block" v-model="selected" @blur="onBlur" @focus="onFocus" :class="{ error: hasError }">
        <option value="AL">Alabama</option>
        <option value="AK">Alaska</option>
        <option value="AZ">Arizona</option>
        <option value="other">Other/Non-USA</option>
      </select>
    </div>
    <div class="col-3fr">
      <input v-show="selected === 'other'" class="full-width-block" type="text" placeholder="Please specify" v-model="other" @blur="onBlur" @focus="onFocus" :class="{ error: hasError }" />
    </div>
  </div>
</template>

<script>
  let blurTimeout;

  export default {
    props: ['value', 'hasError'],
    data() {
      return {
        selected: null,
        other: null,
      };
    },
    computed: {
      state() {
        return this.selected === 'other' ? this.other : this.selected;
      },
    },
    watch: {
      state(val) {
        this.$emit('input', val);
      },
    },
    methods: {
      onBlur() {
        blurTimeout = window.setTimeout(() => {
          this.$emit('blur');
        }, 50);
      },
      onFocus() {
        window.clearTimeout(blurTimeout);
      },
    },
  };
</script>
```

The blur stuff is to support vee-validate, and the input event is to support v-model. It works well in coolest-platform, could be adapted for this to simplify.

src/events/cd-event-add-child-ticket.vue
heyitsmaimai: I tried that as well. The issue for the if/show is that the gender has to be equal to the value 'specify' and when the input box shows it already has 'specify' in it as the gender. once i try to change it to anything else the box won't show (as the gender will not be equal to 'specify'). so genderSelect and genderInput are used and then gender is computed from there

src/events/cd-event-add-child-ticket.vue
heyitsmaimai: I just changed it to another paragraph element outside that error instead of using the span. If i use the span it inherits the text-danger class which is not nice

src/events/cd-event-add-child-ticket.vue
Wardormeur: I guess onTouch you could reset the value of gender from specify to null ?

heyitsmaimai: But the input field shows based on the condition that the value of gender is specify, so if it is reset then it wouldn't show at all

test/unit/specs/events/cd-event-add-child-tickets.spec.js
```diff
@@ -229,7 +166,7 @@ describe('Add Child Ticket', () => {
         vm.tickets = [{
           approvedApplications: 0,
           deleted: 0,
-          entity$: 'ticketentity',
+          entity: 'ticketentity',
```
Wardormeur: err. By remove, i meant, remove the entity$ field alltogether, not just the $ :p

src/common/cd-gender-component.vue
```diff
@@ -0,0 +1,58 @@
...
+<div>
+  <select class="select-box form-control" data-vv-name="gender" data-vv-validate-on="blur" v-validate="'required'" v-model="genderSelect">
+    <option value="" selected data-default disabled>{{ $t('Select an option') }}</option>
+    <option value="male">{{ $t('Male') }}</option>
+    <option value="female">{{ $t('Female') }}</option>
+    <option value="prefer not to answer">{{ $t('Prefer not to answer') }}</option>
+    <option value="specify">{{ $t('Specify Identity') }}</option>
+  </select>
+  <input v-show="specifyGender" class="form-control" v-model="genderInput" :placeholder="$t('Identify as...')" data-vv-name="gender" data-vv-validate-on="blur" v-validate="specifyGender ? 'required' : ''"/>
+<p class="gender-err text-danger" v-show="errors.has('gender:required')">{{ $t('Gender is required') }}<br/><a v-on:click="showWhy">{{ $t('Why is this required? Click here to find out more') }}</a></p>
```
Wardormeur: Where is the indentation gone? :fox_face:

heyitsmaimai: woops, will fix that

src/common/cd-gender-component.vue
```diff
@@ -1,32 +1,34 @@
...
-<p class="gender-why" v-show="whyGender && errors.has('gender:required')">{{ $t(`We want to provide activities that appeal to people regardless of their gender.`) }}<br/>{{ $t(`To check how well we are succeeding, we'd like to find out whether or not people of different genders are equally likely to take part.`) }}</p>  
+  <input v-show="specifyGender" class="form-control" v-model="genderInput" :placeholder="$t('Identify as...')" @blur="onBlur" @focus="onFocus"/>
 </div> 
 </template>
 
 <script>
   export default {
     name: 'cd-gender-component',
-    props: ['gender'],
+    props: ['genders'],
```
heyitsmaimai: ignore this lol

test/unit/specs/events/cd-event-add-child-tickets.spec.js
```diff
@@ -39,6 +39,24 @@ describe('Add Child Ticket', () => {
     });
   });
 
+  // put test for filter
```
heyitsmaimai: this does not work. i don't know why i pushed this. will fix

src/events/cd-event-sessions.vue
```diff
@@ -78,22 +67,31 @@
...
         const validatedChildComponents = await this.validateAllChildComponents();
-        validatedChildComponents.forEach((childIsValid) => {
-          if (!childIsValid) {
-            this.valid = false;
-          } else {
-            this.valid = true;
-          }
-        });
+        if (validatedChildComponents.length > 0) {
+          validatedChildComponents.forEach((childIsValid) => {
```
Wardormeur: The value of valid is overwritten by the last one. If the last one is valid, everything will be considered as valid.
Removing the if/else and returning `this.valid = this.valid && childIsValid` should be enough.

src/events/cd-event-sessions.vue
```diff
@@ -104,45 +102,54 @@
...
-        const validatedChildComponents = await this.validateAllChildComponents();
-        validatedChildComponents.forEach((childIsValid) => {
-          if (!childIsValid) {
-            this.valid = false;
-          } else {
-            this.valid = true;
-          }
+      async addNewChildren() {
+        let promiseChain = Promise.resolve();
+        (this.$refs.allChildComponents).forEach((child) => {
```
Wardormeur: Can be simplified without the PromiseChain :)
Give a try to Promise.all

src/events/cd-event-sessions.vue
```diff
@@ -104,45 +102,54 @@
...
+        this.errors.clear();
+        this.valid = false;
+        await this.checkValidatedChildComponents();
 
-        const validPhone = await this.$validator.validateAll();
+        if (this.showPhone) {
+          this.validPhone = await this.$validator.validateAll();
+        }
 
         if (this.valid === false) {
```
Wardormeur: To avoid the if/else/elseif, define either your default to be the positive result (await this.submitForm) or "any other scenario", which means negating the positive scenario

src/events/cd-event-sessions.vue
```diff
@@ -78,22 +67,31 @@
       totalBooked() {
         return this.children.length;
       },
+      applications() {
+        return [].concat(...(this.children).map(child => child.value));
```
Wardormeur: map already returns an array, no need to concat. If it doesn't (because this.children is empty?) you can safely default it to `map() || []`

src/events/cd-event-sessions.vue
```diff
@@ -62,12 +51,12 @@
...
         valid: true,
+        validPhone: true,
         user: {},
         phone: '',
         submitted: false,
       };
     },
-    // needs tests for these bois
     computed: {
       showPhone() {
```
Wardormeur: What's the idea behind showPhone? Is it to display only for parents rather than o13 ?
The current condition basically forces it to anyone which doesn't have a phone number, which I don't understand

src/events/cd-event-sessions.vue
heyitsmaimai: I think `this.valid = childIsValid` is even better bc it uses the latest value :) with the  `this.valid && childIsValid`, if the previous value of this.valid is false (say they forgot their ticket) and then they fix that issue, it will still give a false value (false&&true). thanks

src/events/cd-event-sessions.vue
heyitsmaimai: Ok will try that

src/events/cd-event-sessions.vue
Wardormeur: Hum, let me rephrase.
In the current scenario, if your first child is invalid but your second child is, this.valid will be true, which is wrong.
The validity upon checking should be resetted at the beginning of this function to ensure that, or I don't get the point of this function.
Technically speaking, this function could be summed up as `validatedChildComp.every((childIsValid) => childIsValid))`

src/events/cd-event-sessions.vue
Wardormeur: Would https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/entries help then instead ?

src/events/cd-event-sessions.vue
Wardormeur: So it's more a retro-compatibility thingy more than anything else for previous parents (returning users) for which the phone number is missing? cc @conoramurphy 
In which case, it can be simplified into `return !!this.user.phone`

src/events/cd-event-sessions.vue
heyitsmaimai: Oh ok gotcha!! Will implement the .every in this function

src/events/cd-event-sessions.vue
heyitsmaimai: yea exacty

src/events/cd-event-sessions.vue
heyitsmaimai: I don't think so from what i have tried with it

src/events/cd-event-add-child-ticket.vue
```diff
@@ -185,6 +185,60 @@
...
       padding: 8px 0px 24px;
-    }                   
+      min-height: 32px;
+    }
+     &__ticket-selector .multiselect {
+      font-family: Lato, Arial, Helvetica, sans-serif;
+      font-size: 14px;
+      min-height: 32px;
+      border: 1px solid #ccc;
+      border-top-color: rgb(204, 204, 204);
```
Wardormeur: clean up the borders, it's all the sames

src/events/cd-event-add-child-ticket.vue
```diff
@@ -185,6 +185,60 @@
...
+.cd-child-ticket__ticket-selector .multiselect__tags {
+  min-height: 32px;
+  padding: 2px 40px 0 2px;
+}
+.cd-child-ticket__ticket-selector .multiselect__tags .multiselect__tags-wrap .multiselect__tag {
+  margin-bottom: 0px;
+  margin-top: 1px;
+}
+.cd-child-ticket__ticket-selector .multiselect__tags .multiselect__single{
+ font-size: 14px;
```
Wardormeur: use the proper font size : https://github.com/CoderDojo/cp-zen-frontend/blob/master/src/common/variables.less

src/events/cd-event-add-child-ticket.vue
```diff
@@ -185,6 +185,60 @@
     &__ticket-selector {
       max-width: 50%;
       padding: 8px 0px 24px;
-    }                   
+      min-height: 32px;
+    }
+     &__ticket-selector .multiselect {
+      font-family: Lato, Arial, Helvetica, sans-serif;
```
Wardormeur: Use the proper variable instead https://github.com/NateWr/bootstrap3/blob/master/bootstrap/less/variables.less#L45

src/events/cd-event-sessions.vue
```diff
@@ -18,12 +18,12 @@
...
-    <p class="text-danger" v-show="showPhone && errors.has('phone:regex')">* Please include the country code. For example, a phone number in Ireland should begin +353</p>
+    <p class="text-danger" v-show="showPhone && errors.has('phone:regex')">* Please include the plus symbol(+) and country code. For example, a phone number in Ireland should begin +353</p>
 
     <div class="cd-event-sessions__next-block">
       <p v-show="totalBooked <= 0" class="cd-event-sessions__next-ticket-select-error text-danger"> {{ $t('Please select at least one ticket') }}</p>
       <p v-show="errors.has('submitChildComponentsFailed')" class="cd-event-sessions__next-child-error text-danger">There was a problem confirming tickets: A ticket is not valid/complete. Please correct this and try again.</p>
-      <button class="cd-event-sessions__next btn btn-primary" tag="button" @click="next">Request booking for {{totalBooked}} tickets</button>
+      <button class="cd-event-sessions__next btn btn-primary" tag="button" @click="next">Request booking for {{totalBooked}} ticket(s)</button>
```
Wardormeur: Isn't "request booking" supposed to be used only for event with approval?

src/events/cd-event-add-child-ticket.vue
heyitsmaimai: just seeing this, i agree

src/events/cd-event-sessions.vue
```diff
@@ -167,21 +173,9 @@
...
-      margin: 4px 0;
-    }
-    &__session {
-      border-style: solid;
-      border-color: @cd-orange;
-      border-width: 1px 1px 3px 1px;
-      padding: 16px;
-      margin-bottom: 24px;
-    }
+    // &__description {
```
Wardormeur: Commented out code, bin it if it's not used :)

src/events/cd-event-add-child-ticket.vue
```diff
@@ -188,38 +189,20 @@
       min-height: 32px;
     }
      &__ticket-selector .multiselect {
-      font-family: Lato, Arial, Helvetica, sans-serif;
-      font-size: 14px;
+      font-family: Lato, @font-family-sans-serif;
+      font-size: @font-size-base;
       min-height: 32px;
       border: 1px solid #ccc;
-      border-top-color: rgb(204, 204, 204);
```
Wardormeur: You removed all the borders, is that alright ?

src/events/cd-event-sessions.vue
```diff
@@ -67,6 +67,12 @@
       applications() {
         return [].concat(...(this.children).map(child => child.value));
       },
+      bookingType() {
```
Wardormeur: Can be simplified as a ternary

test/unit/specs/events/cd-event-add-child-tickets.spec.js
```diff
@@ -0,0 +1,314 @@
...
+          sessionId: 'sessionId',
+          totalApplications: 0,
+          type: 'ninja' }];
+        vm.event = { dojoId: 'dojoId' };
+
+        // ACT
+
+        // ASSERT
+        expect(vm.applications).to.deep.equal([{
+          created: new Date(),
```
Wardormeur: This is very likely what is failing your tests unless you use something like http://sinonjs.org/releases/v1.17.7/fake-timers/

Wardormeur:
Flaky e2e tests

PR #125, opened 2018-05-24
Fix recurring event for past event display

Wardormeur:


Wardormeur:
Well. Yeah
Offer welcome for the text-change, I didn't do it because my first concern was to have the event display and not have an empty category on the page (because the previous event category properly display)

DanielBrierton:
I think I'd be inclined to just remove any info about it being recurring. It's a past event, so does it really matter? You care about the last time a Dojo ran, not the x times before that too.

Wardormeur:
So.. "Last date"/"Last time" ?

PR #128, opened 2018-05-28
Change 'add youth ticket' to 'add another youth ticket'

heyitsmaimai:


Wardormeur:
You probably want to remove the "Jumpy All First Letter Capital" syndrom that the web has

PR #130, opened 2018-05-30
Special requirements option

heyitsmaimai:


src/common/cd-special-req-component.vue
```diff
@@ -0,0 +1,67 @@
+<template>
+<div>
+  <select class="select-box form-control" v-model="specialReqSelect" @blur="onBlur" @focus="onFocus">
+    <option value="" selected data-default disabled>{{ $t('Select an option') }}</option>
+    <option value="no">{{ $t('No') }}</option>
+    <option value="yes">{{ $t('Yes') }}</option>
+  </select>
+  <input v-show="selectYes" class="form-control" v-model="specialReqInput" :placeholder="$t('Type here')" @blur="onBlur" @focus="onFocus"/>
```
Wardormeur: Could benefit from something more specific than "Type here" like "will be late" or "allergic to kids"

src/common/cd-special-req-component.vue
```diff
@@ -0,0 +1,67 @@
+<template>
+<div>
+  <select class="select-box form-control" v-model="specialReqSelect" @blur="onBlur" @focus="onFocus">
+    <option value="" selected data-default disabled>{{ $t('Select an option') }}</option>
```
Wardormeur: Default to "No" as the textbox is hidden by default anyway

src/common/cd-special-req-component.vue
```diff
@@ -0,0 +1,67 @@
...
+        this.blurTimeout = window.setTimeout(() => {
+          this.$emit('blur');
+        }, 50);
+      },
+      onFocus() {
+        window.clearTimeout(this.blurTimeout);
+      },
+    },
+    computed: {
+      selectYes() {
```
Wardormeur: "hasSpecialReq" instead?

src/common/cd-special-req-component.vue
```diff
@@ -0,0 +1,67 @@
...
+        return this.specialReqSelect === 'yes';
+      },
+      specialRequirement: {
+        get() {
+          if (this.specialReqSelect === 'yes') {
```
Wardormeur: use the computed prop instead ?

src/common/cd-special-req-component.vue
```diff
@@ -0,0 +1,67 @@
...
+            return this.specialReqInput;
+          }
+          return this.specialReqSelect;
```
Wardormeur: return a falsy value instead, like null or undefined?

Wardormeur: The only value you want out of this component is "the text" or nothing

src/common/cd-special-req-component.vue
```diff
@@ -0,0 +1,67 @@
...
+        },
+        set(specialReqVal) {
+          if (specialReqVal === 'yes') {
+            this.specialReqSelect = 'yes';
+            this.specialReqInput = specialReqVal;
```
Wardormeur: Why presetting the value of specialReqInput with "yes" ?

src/events/cd-event-add-child-ticket.vue
```diff
@@ -32,6 +32,10 @@
         <multiselect v-model="tickets" :options="childTickets" group-label="name" group-values="tickets" :multiple="true" :searchable="false" :group-select="false" :placeholder="$t('Select Event Tickets')" track-by="id" label="name" @close="onBlur" @open="onFocus" :data-vv-name="`tickets-${id}`" v-validate="'required'"></multiselect>
       </div>
       <p class="cd-child-ticket__ticket-select-err text-danger" v-show="errors.has(`tickets-${id}:required`)">{{ $t('Ticket selection is required') }}</p>  
+
+      <label>{{ $t('Special Requirements') }}</label>
+      <special-req-component class="cd-child-ticket__gender-selector" v-model="specialRequirement" data-vv-value-path="value" data-vv-name="specialRequirement" v-validate="'required'"></special-req-component>
+      <p class="cd-event-ticket__special-req-err text-danger" v-show="errors.has('specialRequirement:required')">{{ $t('Special requirements is required') }}</p>
```
Wardormeur: Not sure this should be required

src/events/cd-event-add-child-ticket.vue
```diff
@@ -117,7 +124,7 @@
           dojoId: this.event.dojoId,
           ticketId: ticket.id,
           userId: this.userId,
-        }));
+        }, this.specialRequirement === 'no' ? '' : { specialRequirement: this.specialRequirement })));
```
Wardormeur: If you return a falsy value when not set, you wouldn't have to compare to a string ;)

src/common/cd-special-req-component.vue
heyitsmaimai: How is this? 'Please type any special requirements you have here (ex. Need wheelchair access)'

src/common/cd-special-req-component.vue
heyitsmaimai: yeah sorry i did change that if statement so it's not really correct any longer. This is for loading the data in say when you are coming back to edit etc. so really, the if statement should be testing whether specialReqVal even exists and then setting the corresponding values to whatever they are. does that make sense?

src/common/cd-special-req-component.vue
Wardormeur: Better :) @conoramurphy does that fit ?

src/common/cd-special-req-component.vue
Wardormeur: It's a dropdown, not an alien web component :p

src/common/cd-special-req-component.vue
Wardormeur: Yep

src/common/cd-special-req-component.vue
```diff
@@ -0,0 +1,64 @@
...
+      specialRequirement: {
+        get() {
+          if (this.hasSpecialReq) {
+            return this.specialReqInput;
+          }
+          return this.specialReqSelect;
+        },
+        set(specialReqVal) {
+          if (!specialReqVal) {
+            this.specialReqSelect = 'yes';
```
Wardormeur: shouldn't that be the opposite?
When specialReq is set, it's a "yes" and the input is filled.
When specialReq is not set, it's a no, which means resetting the input to null and resetting the select to empty

heyitsmaimai: yes you are correct my bad

src/common/cd-special-req-component.vue
conoramurphy: Just to keep consistent with how we talk elsewhere I'd go with.

`Please tell us about any special requirements that we might need to know of here (eg. Need wheelchair access)`

heyitsmaimai: okeeeeee

src/common/cd-special-req-component.vue
```diff
@@ -0,0 +1,59 @@
...
+  <a v-show="!this.hasSpecialReq" class="cd-special-req-component_add" @click="showSpecialReq"> <span><i class="fa fa-plus-circle" aria-hidden="true"></i></span> <span>{{ $t('Add special requirements') }}</span></a>
+  <label v-show="this.hasSpecialReq">{{ $t('Special requirements') }}</label>
+  <textarea v-cols="50" v-show="this.hasSpecialReq" class="form-control" v-model="specialReqInput" :placeholder="$t('Please tell us about any special requirements that we might need to know of here (eg. need wheelchair access)')" @blur="onBlur" @focus="onFocus"/>
+</div> 
+</template>
+
+<script>
+  export default {
+    name: 'cd-special-req-component',
+    props: ['value'],
```
Wardormeur: unused prop?

src/common/cd-special-req-component.vue
```diff
@@ -0,0 +1,59 @@
+<template>
+<div>
+  <a v-show="!this.hasSpecialReq" class="cd-special-req-component_add" @click="showSpecialReq"> <span><i class="fa fa-plus-circle" aria-hidden="true"></i></span> <span>{{ $t('Add special requirements') }}</span></a>
+  <label v-show="this.hasSpecialReq">{{ $t('Special requirements') }}</label>
```
Wardormeur: You can wrap the label and the text-area to avoid 2 v-show and make clear the "states" in the template

src/common/cd-special-req-component.vue
```diff
@@ -0,0 +1,59 @@
...
+    },
+    watch: {
+      specialRequirement() {
+        this.$emit('input', this.specialRequirement);
+      },
+    },
+  };
+</script>
+
+<style scoped lang="less">
```
Wardormeur: not used? remove :)

test/unit/specs/common/cd-special-req-component.spec.js
```diff
@@ -0,0 +1,125 @@
...
+
+        // ASSERT
+        expect(window.clearTimeout).to.have.been.calledOnce;
+        expect(window.clearTimeout).to.have.been.calledWith('foo');
+      });
+    });
+  });
+
+  describe('computed', () => {
+    describe('computed.hasSpecialReq', () => {
```
Wardormeur: hasSpecialReq is not computed

test/unit/specs/common/cd-special-req-component.spec.js
```diff
@@ -0,0 +1,125 @@
...
+
+        // ACT
+
+        // ASSERT
+        expect(vm.hasSpecialReq).to.equal(false);
+      });
+    });
+
+    describe('computed.specialRequirement', () => {
+      it('should return the input requirement value if the if yes was selected', async () => {
```
Wardormeur: if the if ! :)

Wardormeur: Also, still using the 'yes' dropdown value ?

test/unit/specs/common/cd-special-req-component.spec.js
heyitsmaimai: sorry didnt get to tests yet, was just about to fix em

test/unit/specs/common/cd-special-req-component.spec.js
heyitsmaimai: all doneeeee

src/common/cd-special-req-component.vue
```diff
@@ -0,0 +1,55 @@
...
+
+</div> 
+</template>
+
+<script>
+  export default {
+    name: 'cd-special-req-component',
+    data() {
+      return {
+        specialReqSelect: '',
```
Wardormeur: deeeeead

test/unit/specs/common/cd-special-req-component.spec.js
```diff
@@ -0,0 +1,111 @@
...
+  beforeEach(() => {
+    sandbox = sinon.sandbox.create();
+  });
+
+  afterEach(() => {
+    sandbox.restore();
+  });
+
+  describe('methods', () => {
+    describe('methods.onBlur()', () => {
```
Wardormeur: There is no onBlur anymore

test/unit/specs/common/cd-special-req-component.spec.js
```diff
@@ -0,0 +1,111 @@
...
+        // ASSERT
+        expect(window.setTimeout).to.have.been.calledOnce;
+        expect(window.setTimeout).to.have.been.calledWith(sinon.match.func, 50);
+        expect(vm.$emit).to.have.been.calledOnce;
+        expect(vm.$emit).to.have.been.calledWith('blur');
+        expect(vm.blurTimeout).to.equal('foo');
+      });
+    });
+
+    describe('methods.onFocus()', () => {
```
Wardormeur: No onFocus anymore

test/unit/specs/common/cd-special-req-component.spec.js
heyitsmaimai: oh woop will fix

PR #124, opened 2018-05-22
Booking flow/add event full restriction

Wardormeur:
TODO : current booking restriction (1 ticket left, 2 bookers)

PR #134, opened 2018-06-05
Bugfix/confirmation page set new user

Wardormeur:
Need cherry-pick after https://github.com/CoderDojo/cp-zen-frontend/pull/124 is merged

### DIFF
--- a/src/events/order/cd-booking-confirmation.vue
+++ b/src/events/order/cd-booking-confirmation.vue
@@ -1,6 +1,5 @@
 <template>
-  <div class="cd-booking-confirmation">
-
+  <div class="cd-booking-confirmation" v-if="order && event">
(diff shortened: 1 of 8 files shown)


src/events/order/cd-booking-confirmation.vue
```diff
@@ -36,7 +36,7 @@
           <span class="cd-booking-confirmation__booking-details-box-title">{{ $t('Location') }}</span>
         </div>
         <div class="cd-booking-confirmation__booking-details-box-content">
-          <div class="cd-booking-confirmation__event-location">
+          <div class="cd-booking-confirmation__event-location" v-if="event.city && event.country">
```
DanielBrierton: Why the `v-if` here? Shouldn't every event have a city and country?

PR #137, opened 2018-06-11
Fix and detect missing translations

Wardormeur:
without corresponding value in cp-translation
Relies on e2e tests and browser logs

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,6 +9,7 @@
     "build": "node build/build.js",
     "unit": "cross-env BABEL_ENV=test karma start test/unit/karma.conf.js --single-run",
     "e2e": "wdio wdio.conf.js",
+    "i18n": "I18N_DEBUG=true yarn e2e-with-mocks",
     "e2e-with-mocks": "concurrently --kill-others --success first \"yarn mock-server > /dev/null\" \"TARGET_URL_BASE=http://localhost:8080/ wdio wdio.conf.js\" \"yarn start\" ",
     "start-with-mocks": "concurrently --kill-others --success first \"yarn mock-server\" \"yarn start\" ",
     "test": "yarn unit && yarn e2e-with-mocks",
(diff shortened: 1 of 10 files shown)


Wardormeur:
relies on https://github.com/CoderDojo/cp-translations/pull/62

src/events/cd-user-ticket-list-item.vue
```diff
@@ -24,7 +24,7 @@
         <a :href="event.eventbriteUrl | cdUrlFormatter" target="_blank" class="btn btn-lg btn-primary cd-user-ticket-list-item__view">{{ $t('See Details and Book') }}</a>
       </div>
       <div v-if="hasApplications" class="cd-user-ticket-list-item__view-applications">
-        <h4 class="cd-user-ticket-list-item__view-application-title"><i class="fa fa-ticket"></i>{{ $t('Tickets') }}</h4>
+        <h4 class="cd-user-ticket-list-item__view-application-title"><i class="fa fa-ticket"></i>{{ $tc('Tickets') }}</h4>
```
DanielBrierton: Using $tc is useless without

package.json
```diff
@@ -9,6 +9,7 @@
     "build": "node build/build.js",
     "unit": "cross-env BABEL_ENV=test karma start test/unit/karma.conf.js --single-run",
     "e2e": "wdio wdio.conf.js",
+    "i18n": "I18N_DEBUG=true yarn e2e-with-mocks",
```
DanielBrierton: Do we want `I18N_DEBUG` on by default? I guess we probably do?

src/events/cd-user-ticket-list-item.vue
Wardormeur: true, will remove tc and default to plurals

package.json
Wardormeur: It's spammy and very.. hacky right now. I'd prefer us to have a look whenever necssary rather than by default
The best solution would prob be using it as a plugin/reporter, which would allow us to format it better than now.

DanielBrierton: So are you saying you'd rather turn it off? Is it noisy when there _isn't_ missing strings. If not, then the noise is good IMO. Makes sure something is done about the missing strings.

Wardormeur: Well, it doesn't format the output properly because wdio outputs at every end of spec file, instead at every end of test "run" : that's because the memory is lost in-between each spec file. That means you have to scroll back for every spec file running to find the corresponding missing translations : that's crap.
That's why a reporter/plugin is probably needed, and deactivated by default : only show them when you're working & testing a single spec file :)

PR #140, opened 2018-06-13
Add placeholders && transitions while loading data

Wardormeur:


src/events/order/cd-event-sessions.vue
```diff
@@ -220,12 +229,32 @@
 <style scoped lang="less">
   @import "~@coderdojo/cd-common/common/_colors";
   @import "../../common/styles/cd-primary-button";
-
+  @import "../../common/styles/cd-filler-loading";
+  .cd-filler-loading;
   .cd-event-sessions {
+    &--filler {
```
DanielBrierton: All these styles would be the same across all usages of filler, right? Perhaps then a `cd-filler` class can just be exposed by the imported LESS file, and just add that class to the filler div?

Wardormeur: Fair :)

src/common/styles/cd-filler-loading.less
```diff
@@ -1,3 +1,18 @@
...
+    content: " ";
+    position: absolute;
+    height: 100%;
+    width: 100%;
+    animation: fillerLoading .8s linear infinite;
+    background: linear-gradient(to right, transparent 42%, fade(@cd-white, 50%) 50%, transparent 58%) 50% 50%;
+  }
+}
+
 .cd-filler-loading() {
```
DanielBrierton: This doesn't need to be a function anymore, and you definitely don't need to call it in cd-event-sessions.vue anymore. Just import the less file and you should be good!

Wardormeur: :+1:

PR #141, opened 2018-06-13
[DO NOT MERGE] Booking flow/final redirection to booking url

Wardormeur:
Also remove the replace: true for navigation, that swallows the previous page


index.js
```diff
@@ -91,6 +91,57 @@ exports.register = function (server, options, next) {
...
+        }
+      },
+    },
+  });
+  
+  server.route({
+    method: 'GET',
+    path: '/events/{eventId}/confirmation',
+    handler: function (request, reply) {
+      reply(openGraphTemplate({
```
DanielBrierton: Do we really need to bother with open graph on the confirmation page?

Wardormeur: The url is unique, so I'd prefer too

DanielBrierton: We should probably redirect to the sessions page if someone gets to this page and doesn't have an order. We already redirect to LoginOrRegister if they're not logged in.

Wardormeur: Not sure what's a good solution here. A navguard would be nice, but that means injecting the store into it. 
Can a navguard be bypassed if you come from another url, (like, normal flow) and only triggered on "refresh" ?

DanielBrierton: It should be possible. The nav guard gets `from` and `to` passed into it. On refresh, we should be able to check that there is no `from`. Need to check exactly what it is in that scenario, but I don't see how it couldn't work.

src/router/index.js
```diff
@@ -107,7 +109,7 @@ const router = new Router({
           path: 'events/:eventId/confirmation',
           name: 'EventBookingConfirmation',
           component: BookingConfirmation,
-          beforeEnter: loggedInNavGuard,
+          beforeEnter: MultiGuard([orderExistsNavGuard, loggedInNavGuard]),
```
DanielBrierton: Order should be swapped around. Check if they're logged in before trying for the order. Otherwise we could end up requesting the order for user `undefined`

PR #142, opened 2018-06-14
Add a redirection to login-page if the user is already logged-in

Wardormeur:


src/users/cd-login.vue
```diff
@@ -59,6 +59,11 @@
         }
       },
     },
+    async created() {
+      const loggedInUser = (await UserService.getCurrentUser()).body;
```
DanielBrierton: We have a store to get this info from.

Also, could be done as a `beforeEnter` on the route. The first argument, `to`, is the route you're going to, including query params, so you can get the redirect URL from there.

Wardormeur: + replace:true required

PR #144, opened 2018-06-15
Use the store to load the event and the sessions

Wardormeur:


src/events/order/cd-event-sessions.vue
```diff
@@ -69,9 +70,12 @@
     },
     data() {
       return {
+<<<<<<< HEAD
```
DanielBrierton: lulz

src/events/order/cd-event-sessions.vue
```diff
@@ -124,6 +123,7 @@
         }
         return this.existingChildren;
       },
+      ...mapGetters('order', ['event']),
```
DanielBrierton: Huh... I thought we had to set the `store` on each component, but it seems like it inherits from the parent... 👍

src/events/order/cd-event-sessions.vue
Wardormeur: No moar lulz for you

PR #148, opened 2018-06-25
Remove bold from date-picker

Wardormeur:


src/users/cd-create-account.vue
```diff
@@ -336,3 +336,9 @@
     height: 36px;
   }
 </style>
+<style lang="less">
+  @import "~bootstrap/less/variables";
+  .cd-create-account__dob-picker-wrapper .vue-dob-picker label select {
```
DanielBrierton: vue-dob-picker itself doesn't have any styles, so you don't need to override it. Instead, it has a `select-class` prop that is currently set to `form-control`. You can add another class there and override it instead of this. (The bold is coming from Bootstrap, not the DOB picker).

PR #153, opened 2018-06-26
Disallow deletion of last children if it's the only left

Wardormeur:


src/events/order/cd-event-sessions.vue
```diff
@@ -10,7 +10,7 @@
       <ticket v-for="(_user, index) in users" :user="_user" :event="event" :key="_user.userId" :existing-applications="existingApplications[_user.userId]"></ticket>
     </div>
     <!-- TODO : isn't the index enough rather than generating an uuid ? -->
-    <child-ticket v-for="(child, index) in children" ref="allChildComponents" :key="child.id" :eventId="eventId" :event="event" :sessions="sessions" :id="child.id" v-on:delete="deleteChildComponent(index)" v-validate.disable="'submitApplications:required'"></child-ticket>
+    <child-ticket v-for="(child, index) in children" ref="allChildComponents" :key="child.id" :eventId="eventId" :event="event" :sessions="sessions" :id="child.id" v-on:delete="deleteChildComponent(index)" v-validate.disable="'submitApplications:required'" :deletable="users.length > 0 || (users.length <= 0 && children.length > 1)"></child-ticket>
```
DanielBrierton: Checking `users.length <= 0` is kinda pointless since that's the only scenario where we'll get to this check. Just `users.length > 0 || children.length > 1` should suffice.

Wardormeur: Fair, fixed

PR #159, opened 2018-06-28
Maintenance/remove events v2 calls

Wardormeur:
TODO : cancellation related functions


src/users/cd-tickets.vue
```diff
@@ -39,10 +41,21 @@
...
+        return Promise.all(this.usersDojos.map(dojo =>
+          EventService.v3.get(dojo.dojoId, {
+            params: {
+              query,
+              related: 'sessions.tickets',
+            } })))
+          .then((events) => {
+            this.events = (events.reduce((red, dojoEvents) =>
+              red.concat(dojoEvents.body.results), []))
+            .sort(EventUtils.orderByStartTime);
```
DanielBrierton: We don't need this sort anymore, do we? Now that it's handled on the backend.

src/users/cd-tickets.vue
```diff
@@ -39,10 +41,21 @@
...
       async loadEvents() {
-        this.usersDojos.forEach(async (dojo) => {
-          const res = await EventService.loadEvents(dojo.dojoId);
-          this.events = this.events.concat(res.body);
-        }, this);
+        const query = { status: 'published' };
+        query.afterDate = moment().unix();
+        query.utcOffset = moment().utcOffset();
+
+        return Promise.all(this.usersDojos.map(dojo =>
```
DanielBrierton: I'd prefer

src/users/cd-tickets.vue
Wardormeur: We do. It's an aggregation of all Dojos events of which the users belongs to

src/users/cd-tickets.vue
DanielBrierton: I'd prefer
```javascript
return Promise.all(this.usersDojos.map(async (dojo) => {
  const events = await EventService.v3.get(dojo.dojoId, {
    params: {
      query,
      related: 'sessions.tickets',
    }
  });
  this.events = events
    .reduce((red, dojoEvents) => red.concat(dojoEvents.body.results), [])
    .sort(EventUtils.orderByStartTime);
}));
```

DanielBrierton: Basically, I just want to avoid `.then` 😛 It just looks a bit cleaner to me.

Wardormeur: the then is after the promise.all. That would mean that we don't use a reducer but directly concat at each results. It could also work, i guess ?

DanielBrierton: Ah, all the more reason to change it so. 😅

PR #160, opened 2018-06-29
Booking flow visual cleanup

heyitsmaimai:


src/users/cd-create-account.vue
```diff
@@ -208,42 +211,46 @@
...
-      display: block;
-      font-size: 16px;
-      font-weight: bold;
+      margin-bottom: 5px;
+      display: inline-block;
+      font-size: 14px;
+      font-weight: 700;
+    }
+    &__email {
+      max-width: 110%;
```
Wardormeur: Uh, what ?

src/users/cd-create-account.vue
```diff
@@ -258,22 +265,32 @@
       margin-top: 4px;
       font-weight: 300;
     }
-    &__dob-picker {
-      padding-left: 0;
-      max-width: 600px;
+    &__dob-picker-wrapper {
+      max-width: 50%;
```
Wardormeur: Max-width:50% is terrible on mobile

src/users/cd-create-account.vue
heyitsmaimai: oh woop sorry i was playing around with that. will fix

src/users/cd-create-account.vue
heyitsmaimai: why? as in it stays too big or becomes too small?

heyitsmaimai: changed it for 100% on mobile

PR #162, opened 2018-07-04
WIP trying out Cypress for e2e

DanielBrierton:


package.json
```diff
@@ -14,7 +14,10 @@
     "start-with-mocks": "concurrently --kill-others --success first \"yarn mock-server\" \"yarn start\" ",
     "test": "yarn unit && yarn e2e-with-mocks",
     "lint": "eslint --ext .js,.vue src test/unit/specs",
-    "mock-server": "node ./json-server/server.js"
+    "mock-server": "node ./json-server/server.js",
+    "cypress": "concurrently --kill-others --success first \"http-server dist/ --push-state > /dev/null\" \"cypress run\"",
+    "cypress:open": "cypress open",
```
Wardormeur: do we need those commands in the package.json ?

DanielBrierton: `cypress open` is super handy. It watches for changes and re-runs the tests automatically. Plus it shows the tests actually running and allows you to interact with the timeline.

`cypress:install` is also useful to have if you happen to run `yarn` locally rather than the docker container, since the Cypress binaries get installed to `~/.cache/Cypress`. Perhaps worth adding info on this to the README actually. Was intending on doing so.

Wardormeur: Yas :) :+1:

PR #163, opened 2018-07-07
Enhancement/modify booking

Wardormeur:
screenshot from 2018-07-07 12-16-55


Wardormeur:
Change design to use secondary button for cancel tickets

PR #168, opened 2018-08-20
Dashboard children/show children

heyitsmaimai:


test/unit/specs/dashboard/cd-dashboard-children.spec.js
```diff
@@ -0,0 +1,119 @@
...
+  });
+
+  afterEach(() => {
+    sandbox.restore();
+  });
+
+
+  describe('computed', () => {
+    describe('computed.children', () => {
+      it('should return a list containing the current user if isSingle', () => {
```
Wardormeur: Change tests names, isSingle doesn't exists :)

test/unit/specs/dashboard/cd-dashboard-children.spec.js
```diff
@@ -0,0 +1,119 @@
...
+    it('should add a default child if the user is having no child and is not single', async () => {
+      const vm = vueUnitHelper(DashboardChildrenComponentWithMocks);
+      vm.loadCurrentUser = sinon.stub().resolves();
+      vm.loadProfile = sinon.stub().resolves();
+      vm.loadChildren = sinon.stub().resolves();
+      vm.currentUser = null;
+      vm.userProfile = {};
+      vm.userChildren = [];
+
+      await vm.$lifecycleMethods.created();
```
Wardormeur: Missing test? What do we assert here?

Wardormeur:
Also : translations !!1

test/unit/specs/dashboard/cd-dashboard-children.spec.js
heyitsmaimai: errr is it that it actually loads the children if they have kids and doesn't if none?

Wardormeur: It doesn't test anything, it stubs data, but there is no "expect".
You would expect here (and the test name is wrong) that each of the component methods are being called to load the data

src/dashboard/cd-dashboard-children.vue
```diff
@@ -1,16 +1,56 @@
 <template>
   <div class="column">
     <div class="cd-dashboard-children">
-      <h1 class="cd-dashboard-children__header">My Kids</h1>
+      <h1 class="cd-dashboard-children__header">{{ $t('My Children') }}</h1>
+      <div class="cd-dashboard-children__child" v-for="child in children">
+        <h3>{{ child.firstName }} {{ child.lastName }}</h3>
```
DanielBrierton: Oh, actually... I think we have `child.name` available that has them merged already.

src/dashboard/cd-dashboard-children.vue
```diff
@@ -24,9 +64,16 @@
     margin-left: auto;
     min-height: 100%;
     display:flex;
+    flex-direction: column;
```
DanielBrierton: Sorry for third random review, just noticed this in a different PR. 😅  Seems like there's no need for this being flex now that it's column.

Would also be curious if `min-height` is needed here too. I think it is for the moment, and I think I know why, but we should be able to fix that to simplify things. Don't worry about it for now. I'll probably do an overall audit of the CSS anyway and see where we can simplify things. We'll see. 😛

heyitsmaimai: oh the flex column is for the elements inside the children to be displayed stacked 
-
-
-
^ like that

DanielBrierton: Column stacking is the default behaviour for block elements, which is all that's used inside. It's no biggie, but it should work without flex unless I'm missing something.

heyitsmaimai: yeah all good. just fixed that there

PR #170, opened 2018-08-23
Dashboard children/edit children

heyitsmaimai:
depends on [cp-zen-frontend#168](https://github.com/CoderDojo/cp-zen-frontend/pull/168)

### DIFF
--- a/src/dashboard/cd-dashboard-children.vue
+++ b/src/dashboard/cd-dashboard-children.vue
@@ -118,7 +118,6 @@
 
     &__edit-child {
       font-size: 14px;
-      display: flex;
       align-self: flex-end;
       padding: 0 0 0 5px;
     }


src/dashboard/cd-dashboard-children.vue
```diff
@@ -23,10 +69,58 @@
...
+        margin-bottom: 8px;
+      }
+      &-text {
+        text-align: center;
+      }
+    }
+
+    &__edit-child {
+      font-size: 14px;
+      display: flex;
```
DanielBrierton: This element doesn't need to be `display: flex;` as you're not distributing any of it's contents. `align-self` works fine because this is a child of a flex element.

heyitsmaimai: gotcha

PR #171, opened 2018-08-23
Dashboard children/loading

heyitsmaimai:
depends on [cp-zen-frontend#168](https://github.com/CoderDojo/cp-zen-frontend/pull/168)  [cp-zen-frontend#169](https://github.com/CoderDojo/cp-zen-frontend/pull/169)  [cp-zen-frontend#170](https://github.com/CoderDojo/cp-zen-frontend/pull/170). Needs to be edited when these are merged

### DIFF
--- a/src/dashboard/cd-dashboard-children.vue
+++ b/src/dashboard/cd-dashboard-children.vue
@@ -3,7 +3,10 @@
     <div class="cd-dashboard-children">
       <h1 class="cd-dashboard-children__header">{{ $t('My Children') }}</h1>
       <div class="cd-dashboard-children__child" v-for="child in children">
-        <h3 class="cd-dashboard-children__name">{{ child.name }}</h3>
+        <h3 class="cd-dashboard-children__name">
+          {{ child.name }}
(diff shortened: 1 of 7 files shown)


src/dashboard/cd-dashboard-children.vue
```diff
@@ -1,32 +1,151 @@
...
+          this.userChildren = (await Promise.all(
+            this.userProfile.children.map(
+              child => UserService.userProfileData(child))))
+            .map(res => res.body);
+        }
+      },
+    },
+    async created() {
+      await this.loadProfile();
+      this.loadChildren().then(this.loadedChildren = true);
```
DanielBrierton: I'm not sure how this isn't throwing some sort of syntax error. I didn't think you could pass an assignment into something that's expecting a function, but I guess that's just JavaScript for you! 😅 Pretty sure this will end up assigning loadedChildren to true before children have actually loaded.

Best thing to do here is either move `this.loadedChildren = true` to the `loadChildren` function, or here in `created` do:
``` javascript
await this.loadChildren();
this.loadedChildren = true;
```

PR #172, opened 2018-08-27
Dashboard news/forums / set up service

heyitsmaimai:


src/dashboard/service.js
```diff
@@ -0,0 +1,14 @@
+import Vue from 'vue';
+
+const NewsForumsService = {
+  // load recent forums
+  loadForums() {
+    return Vue.http.get(`${Vue.config.forumsUrlBase}/api/recent/new`);
```
DanielBrierton: Ideally, for both this and WP, we'd limit the results to six, since that's the most we'll display (at the moment anyway). That said, documentation is a bit... lacking... and I've been unable to figure out how. 😅 Did you look at all? Perhaps some fresh eyes might spot something I missed.

src/dashboard/service.js
```diff
@@ -0,0 +1,14 @@
+import Vue from 'vue';
+
+const NewsForumsService = {
```
DanielBrierton: I'm not mad on this service name... `CommunityUpdatesService` maybe? Not a big fan of that either, but prefer it... Any other ideas?

src/dashboard/service.js
heyitsmaimai: Yeah so, only thing I can think of looking at the documentation is returning a modified response with 6 results? https://github.com/pagekit/vue-resource#example ?? I dunno if that would be a good way to do that though so

DanielBrierton: I mean on the docs for wp-json and NodeBB. I've found the relevant docs on wp-json now (I obviously didn't look very far 😅). Just add a `per_page` GET param - https://developer.wordpress.org/rest-api/using-the-rest-api/pagination/

NodeBB though, I haven't been able to find the equivalent...

PR #173, opened 2018-08-28
Dashboard news forums/merge posts

heyitsmaimai:
depends on [#172](https://github.com/CoderDojo/cp-zen-frontend/pull/172)


src/dashboard/cd-dashboard-news.vue
```diff
@@ -7,8 +7,45 @@
...
     name: 'cd-dashboard-news',
+    data() {
+      return {
+        news: null,
+        forums: null,
+      };
+    },
+    computed: {
+      allPosts() {
+        return (this.news).map(post => ({
```
DanielBrierton: Since this computed property gets a bit more complex with further PRs, I think it would be best to split it up into a few props.

I'm struggling to come up with good names. All I can think of are `transformedNews`, `transformedForumPosts` and `allUpdates` or something like that...

src/dashboard/cd-dashboard-news.vue
```diff
@@ -7,8 +7,45 @@
...
+        const res = await NewsForumsService.loadNews();
+        this.news = res.body;
+      },
+      async loadForums() {
+        const res = await NewsForumsService.loadForums();
+        this.forums = res.body.topics;
+      },
+    },
+    async created() {
+      await this.loadNews();
```
DanielBrierton: There's actually no need to `await` these. They can load in parallel without any issue. You'll generally only await when you need a result from the first function call in the second one, which isn't the case here.

src/dashboard/cd-dashboard-news.vue
DanielBrierton: In addition, instead of using `arr1.concat(arr2)`, nowadays you can just do `[...arr1, ...arr2]`. Either way works fine, and I'm not that picky. Just more of an FYI.

src/dashboard/cd-dashboard-children.vue
```diff
@@ -14,6 +17,13 @@
         <p class="cd-dashboard-children__badges-none" v-else>{{ $t('{name} doesn\'t have any badges yet. Talk to the organisers of your Dojo to learn how {name} can be rewarded through badges.', { name: child.firstName }) }}</p>
       </div>
     </div>
+    <div v-else class="cd-dashboard-children">
```
DanielBrierton: Seems like there's a slight issue with this PR. Not quite sure why, but these changes shouldn't be here since they're already in master... 😅

src/dashboard/cd-dashboard-news.vue
heyitsmaimai: Yeah that makes sense. how about formattedNews/Forums ??

src/dashboard/cd-dashboard-children.vue
heyitsmaimai: Yeah I will rebase off master and fix dat

src/dashboard/cd-dashboard-news.vue
DanielBrierton: Yeah, that sounds good

src/dashboard/cd-dashboard-children.vue
heyitsmaimai: Ok i did not fix that

heyitsmaimai: Do I have to force push? @DanielBrierton

PR #174, opened 2018-08-29
Dashboard news forums/show dates types titles

heyitsmaimai:
depends on [#173](https://github.com/CoderDojo/cp-zen-frontend/pull/173)


src/dashboard/cd-dashboard-news.vue
```diff
@@ -1,14 +1,77 @@
...
+            link: post.link,
+            title: post.title.rendered,
+          })).concat((this.forums).map(post => ({
+            type: 'Forums',
+            date: (post.timestampISO.split('T')[0]).split('-').join(''),
+            link: `https://forums.coderdojo.com/topic/${post.slug}`,
+            title: post.title,
+          }))));
+
+          return this.sortPostsByDate(joinedPosts).splice(0, 6);
```
DanielBrierton: If the `map`s are moved into their own properties as per my review in #173, then we could probably just move the sort into this computed property instead.

That said, there's nothing really wrong with having a method that does the sort separately either. I guess make the call when `allPosts` is simplified a bit?

src/dashboard/cd-dashboard-news.vue
```diff
@@ -1,14 +1,77 @@
...
+      async loadForums() {
+        const res = await NewsForumsService.loadForums();
+        this.forums = res.body.topics;
+      },
+      sortPostsByDate(posts) {
+        const sortedPosts = posts.sort((a, b) =>
+         a.date - b.date);
+        return (sortedPosts.map(post => (Object.assign({
+          type: post.type,
+          date: [post.date.slice(0, 4), post.date.slice(4, 6), post.date.slice(6, 8)].reverse().join('/'),
```
DanielBrierton: Two things about this:

1. Since we already map the posts, we should do whatever we need to with the date there.
2. We should probably parse the timestamps as Date objects (or with moment) so we can format based on locale. This works, but it's a little ugly...

src/dashboard/cd-dashboard-news.vue
```diff
@@ -1,14 +1,77 @@
...
+          link: post.link,
+          title: post.title,
+        })))).reverse();
```
DanielBrierton: Given we're writing the comparator for sorting, we shouldn't need to reverse afterwards. Using `b.date - a.date` in the comparator would sort it how we want.

If we parse the timestamps to actual dates though you'll need to adjust the comparator a bit anyway. Thankfully dates can be compared with each other, so it's still pretty straight forward.

src/dashboard/cd-dashboard-news.vue
```diff
@@ -47,6 +63,16 @@
         const res = await NewsForumsService.loadForums();
         this.forums = res.body.topics;
       },
+      sortPostsByDate(posts) {
+        const sortedPosts = posts.sort((a, b) =>
+         b.date - a.date);
+        return (sortedPosts.map(post => (Object.assign({
```
DanielBrierton: Two things here:
1. We could avoid mapping over the posts again by adding a `formattedDate` property to `formattedNews` and `formattedForums`. Really, given the amount of data we have, there's very little performance impact of mapping again, but still...
2. As it stands, there's no need for `Object.assign` here. However, it would be useful in a different way
```javascript
return (sortedPosts.map(post => Object.assign({
  date: post.date.utc().format('DD/MM/YYYY'),
}, post);
```
This will create a new object with the formatted date, and then assign all properties from post to it that don't conflict (so date, for example, will not be copied over).

heyitsmaimai: Gotcha. Was looking into that but I got confused about not overwriting, makes so much sense. thanks

PR #175, opened 2018-08-30
Dashboard news forums/loading

heyitsmaimai:
depends on [#174](https://github.com/CoderDojo/cp-zen-frontend/pull/174)


src/dashboard/cd-dashboard-news.vue
```diff
@@ -113,6 +123,30 @@
...
+      }
+
+      &-right{
+        margin: 0 16px 0 16px;
+        max-width: 70%;
+      }
+
+      &--filler {
+        background-color: @cd-very-light-grey;
+        width: 650px;
```
DanielBrierton: This won't play nice with mobile. I _think_ the best course of action is
```css
max-width: 650px;
width: 100%;
```

That should make it not go beyond 650px, but let it shrink.

PR #178, opened 2018-09-04
restrict page to logged in users in router

heyitsmaimai:


cypress/integration/home_spec.js
```diff
@@ -6,6 +6,12 @@ describe('Homepage', () => {
     cy.server();
   });
 
+  it('should redirect to the login page if there is no logged in user', () => {
+    cy.route('/api/2.0/users/instance', 'fx:loggedOutUser');
+    cy.visit('/home');
+    cy.url().should('include', '/login');
```
Wardormeur: Add an alias and ensure the user has been recovered before checking the redirection ?

heyitsmaimai: recovered?

Wardormeur: To ensure that the check to redirect to login is done after the profile (/instance) has been called

PR #179, opened 2018-09-04
Add e2e tests for dashboard news display

Wardormeur:


cypress/fixtures/blogArticles.js
```diff
@@ -0,0 +1,220 @@
+[{
```
DanielBrierton: `.js` but is a JSON file. Change extension or export the array.

Wardormeur: You're right. It happened from c/p from json-server, used node cli to convert it and didn't change the right format afterwards :x

PR #181, opened 2018-09-04
fixup unit tests and get rid of loadedPosts

heyitsmaimai:


Wardormeur:
What happens when you don't have children, and it's normal ? You'll have the animation forever ?

PR #182, opened 2018-09-04
Added latest projects to dashboard

DanielBrierton:


src/common/styles/cd-filler-loading.less
```diff
@@ -8,16 +8,21 @@
...
+    &::before {
+      background: linear-gradient(to right, transparent calc(~"50% - 30px"), #f4f5f6 50%, transparent calc(~"50% + 30px")) 50% 50%;
+    }
   }
 }
 
 @keyframes fillerLoading {
   0% {
-    transform: translate3d(-50%, 0, 0);
+    transform: translate3d(calc(~"-50% - 60px"), 0, 0);
```
Wardormeur: Does this still look OK on SAD ?

src/common/styles/cd-filler-loading.less
```diff
@@ -8,16 +8,21 @@
     position: absolute;
     height: 100%;
     width: 100%;
-    animation: fillerLoading .8s linear infinite;
-    background: linear-gradient(to right, transparent 42%, fade(@cd-white, 50%) 50%, transparent 58%) 50% 50%;
+    animation: fillerLoading .8s ease-in infinite;
+    background: linear-gradient(to right, transparent calc(~"50% - 30px"), fade(@cd-white, 50%) 50%, transparent calc(~"50% + 30px")) 50% 50%;
+  }
+  &--grey-bg {
```
Wardormeur: Couldn't have it be possible to have a function that takes a color ?
I guess we don't have many possible backgrounds, so that may be overkill ?

test/unit/specs/dashboard/cd-dashboard-projects.spec.js
```diff
@@ -0,0 +1,86 @@
...
+
+        // ASSERT
+        expect(vm.isDisplayable).to.deep.equal(false);
+      });
+    });
+  });
+
+  describe('methods', () => {
+    describe('loadProjects', () => {
+      it('should load the current user\'s profile', async () => {
```
Wardormeur: lol message

src/common/styles/cd-filler-loading.less
DanielBrierton: I don't see why it wouldn't. I just made the change so the "shine" is consistent across all fillers. I can check for sure though.

src/common/styles/cd-filler-loading.less
DanielBrierton: There's also the question about how it would look in different colours. The grey bg used in the sidebar just happens to work. Different shades might not.

src/common/styles/cd-filler-loading.less
DanielBrierton: Oh, just copped SAD is in Angular. Needs to be replicated there I guess.

Wardormeur: Nooope, late writing, i meant FTB. SAD doesn't have a loading effect

DanielBrierton: Heh 😅

Specifically tested FTB to be sure, but it's all good!

PR #184, opened 2018-09-05
set up overall dashboard mobile view. update stats overalll view to h…

heyitsmaimai:
…ave same css as children component

### DIFF
--- a/src/dashboard/cd-dashboard-stats.vue
+++ b/src/dashboard/cd-dashboard-stats.vue
@@ -17,16 +17,24 @@
 
 <style scoped lang="less">
   @import "~@coderdojo/cd-common/common/_colors";
+  @import "../common/variables";
 
   .cd-dashboard-stats {
-    background-color: #fff;
+    background-color: #f4f5f6;
     padding: 0 32px;
     margin-left: auto;
     min-height: 100%;
+    max-width: 340px;
     display:flex;
 
     &__header {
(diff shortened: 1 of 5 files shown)


src/dashboard/cd-dashboard.vue
```diff
@@ -54,8 +57,16 @@ export default {
...
-      max-width: 340px;
       flex: 1;
     }
   }
+
+  @media (max-width: @screen-xs-max) {
+    .cd-dashboard {
+
+      &__container {
+        flex-direction: column;
```
Wardormeur: Shouldn't that be column-reverse ?

heyitsmaimai: YES and i can delete the order stuff too nice

src/dashboard/cd-dashboard-stats.vue
```diff
@@ -17,17 +17,25 @@
 
 <style scoped lang="less">
   @import "~@coderdojo/cd-common/common/_colors";
+  @import "../common/variables";
 
   .cd-dashboard-stats {
-    background-color: #fff;
+    background-color: #f4f5f6;
```
Wardormeur: Where does this value comes from ?
Don't we have an eq for that color available ?

heyitsmaimai: It's from the dojo details page - its the left column bar color

PR #185, opened 2018-09-05
Dashboard mobile/children

heyitsmaimai:
need to merge w/ [#184](https://github.com/CoderDojo/cp-zen-frontend/pull/184)

### DIFF
--- a/src/dashboard/cd-dashboard-children.vue
+++ b/src/dashboard/cd-dashboard-children.vue
@@ -145,6 +145,7 @@
       &-image {
         height: 105px;
         width: 100%;
+        object-fit: contain;
         margin-bottom: 8px;
       }
 
@@ -176,6 +177,7 @@
         &-image {
           height: 95px;
           width: 100%;
+          object-fit: contain;
         }
       }
     }
--- a/src/dashboard/cd-dashboard-children.vue
+++ b/src/dashboard/cd-dashboard-children.vue
@@ -3,6 +3,7 @@
     <div v-if="!withoutChildren">
       <div v-if="isDisplayable" class="cd-dashboard-children">
         <h1 class="cd-dashboard-children__header">{{ $t('My Children') }}</h1>
+        <hr class ="cd-dashboard-children__divider visible-xs">
         <div class="cd-dashboard-children__child" v-for="child in children">
           <h3 class="cd-dashboard-children__name">
             {{ child.name }}
@@ -161,6 +162,20 @@
   @media (max-width: @screen-xs-max) {
     .cd-dashboard-children {
       max-width: 100%;
+
+      &__divider {
+        border-color: #CAC4CE; ;
+      }
+
+      &__badge {
+        width: 95px;
+        height: 100%;
+
+        &-image {
+          height: 90px;
+          width: 90px;
+        }
+      }
     }
   }
 </style>
--- a/src/dashboard/cd-dashboard-children.vue
+++ b/src/dashboard/cd-dashboard-children.vue
@@ -85,12 +85,14 @@
 <style scoped lang="less">
   @import "~@coderdojo/cd-common/common/_colors";
   @import "../common/styles/cd-filler-loading";
+  @import "../common/variables";
 
   .cd-dashboard-children {
     background-color: #f4f5f6;
     padding: 0 32px;
     margin-left: auto;
     min-height: 100%;
+    max-width: 340px;
 
     &__header {
       margin: 45px 0 16px 0;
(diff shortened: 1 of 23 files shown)


src/dashboard/cd-dashboard-children.vue
```diff
@@ -155,5 +158,25 @@
       padding: 0 0 0 5px;
     }
   }
+
+  @media (max-width: @screen-xs-max) {
+    .cd-dashboard-children {
+      max-width: 100%;
+
+      &__divider {
+        border-color: #CAC4CE;
```
Wardormeur: Same than other PR, don't we have an existing variable for this color ?

src/dashboard/cd-dashboard-children.vue
```diff
@@ -155,5 +158,25 @@
...
+      }
+
+      &__badge {
+        width: 95px;
+        height: 100%;
+
+        &-image {
+          height: 90px;
```
Wardormeur: We'll have to fix the badges, it's horrible to look at, but I guess we can postpone it to later on (harmonization ticket) ? cc @DanielBrierton

src/dashboard/cd-dashboard-children.vue
```diff
@@ -155,5 +158,25 @@
...
+  @media (max-width: @screen-xs-max) {
+    .cd-dashboard-children {
+      max-width: 100%;
+
+      &__divider {
+        border-color: #CAC4CE;
+      }
+
+      &__badge {
+        width: 95px;
```
Wardormeur: The badges text can be fairly long, why flex them rather than that ?

src/dashboard/cd-dashboard-children.vue
heyitsmaimai: Not that I know of, correct me if i am wrong

src/dashboard/cd-dashboard-children.vue
heyitsmaimai: Ergh yes we do, Daniel mentioned some solution we could use the other day but I dont remember what it was

src/dashboard/cd-dashboard-children.vue
heyitsmaimai: Flex them rather than..? Setting the width manually ?

src/dashboard/cd-dashboard-children.vue
Wardormeur: Flexing the img while giving a size to the parent should work ?

src/dashboard/cd-dashboard-children.vue
Wardormeur: *why not flex them rather than having a set width

src/dashboard/cd-dashboard-children.vue
heyitsmaimai: it was the same color used with the coderdojo.com dividers

src/dashboard/cd-dashboard-children.vue
heyitsmaimai: ahhhh gotcha

src/dashboard/cd-dashboard-children.vue
Wardormeur: Hum. None of them https://github.com/CoderDojo/cd-common/blob/master/common/_colors.less#L61 fits ?

src/dashboard/cd-dashboard-children.vue
heyitsmaimai: errr yes that works but they looks funny with diff sizing .@DanielBrierton what were you sayin we could do the other day??

src/dashboard/cd-dashboard-children.vue
```diff
@@ -169,6 +172,22 @@
...
+        border-color: @divider-grey;
+      }
+
+      &__badge {
+        width: 95px;
+        height: 100%;
+        align-items: center;
+
+        &-image {
+          height: 95px;
```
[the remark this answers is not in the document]
heyitsmaimai: yes they are indeed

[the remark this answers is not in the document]
heyitsmaimai: im going to add a view children button so dont merge yet pls

src/dashboard/cd-dashboard-children.vue
```diff
@@ -18,6 +18,9 @@
           </span>
           <p class="cd-dashboard-children__badges-none" v-else>{{ $t('{name} doesn\'t have any badges yet. Talk to the organisers of your Dojo to learn how {name} can be rewarded through badges.', { name: child.firstName }) }}</p>
         </div>
+        <div v-show="children.length > 3" class="cd-dashboard-children__cta">
+          <a class="cd-dashboard-children__view-all" href="/dashboard/children/">View all children</a>
```
Wardormeur: Translation

PR #188, opened 2018-09-05
Dashboard/projects translations

Wardormeur:


src/projects/service.js
```diff
@@ -1,7 +1,12 @@
 import Vue from 'vue';
 
 const ProjectsService = {
-  list: params => Vue.http.get(`${Vue.config.projectsUrlBase}/api/v1/en/projects`, { params }),
+  list: (language, params) => {
+    // Note : this is a patch due to /projects non-consistent language
+    // eslint-disable-next-line no-unused-vars
```
DanielBrierton: This doesn't seem necessary..?

Wardormeur: Well, the linter was complaining about lang not being used :/

Wardormeur:
Aaaand linter failed. My bad, will fix asap

src/projects/service.js
DanielBrierton: But, like... It is used. Also, conflicts! 😄

Wardormeur: Will have another look at it while solving the conflict, but pretty sure I had the same reaction than you did

DanielBrierton:
Cypress tests seem a little flakey... 😕

PR #190, opened 2018-09-06
add e2e for the view all children link to show when the parent has mo…

heyitsmaimai:
…re than 3 children

### DIFF
--- a/cypress/integration/dashboard/children_spec.js
+++ b/cypress/integration/dashboard/children_spec.js
@@ -59,4 +59,17 @@ describe('Homepage children', () => {
       .should('have.attr', 'href', '/dashboard/children/u1')
       .invoke('text').should('eq', 'See all 3 badges');
   });
+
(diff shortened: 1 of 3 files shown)


cypress/integration/dashboard/children_spec.js
```diff
@@ -59,4 +59,17 @@ describe('Homepage children', () => {
       .should('have.attr', 'href', '/dashboard/children/u1')
       .invoke('text').should('eq', 'See all 3 badges');
   });
+
+  it('should show the view all children link if they have more than 3 children', () => {
```
Wardormeur: Add that it should not show for <3 children (parent1)

heyitsmaimai: on it!

heyitsmaimai: doine

PR #192, opened 2018-09-10
Dashboard/review my children

heyitsmaimai:


cypress/integration/dashboard/children_spec.js
```diff
@@ -1,6 +1,6 @@
 import homePage from '../../pages/home';
 
-describe('Homepage children', () => {
+describe'Homepage children', () => {
```
Wardormeur: running the test would have detected that ;)

heyitsmaimai: i did had an .only there and took it out after i ran it that's why, sorry will fix

PR #195, opened 2018-09-10
Dashboard/review news

heyitsmaimai:


src/dashboard/cd-dashboard-news.vue
```diff
@@ -164,14 +141,11 @@
       }
 
       &-title {
-        margin: 0;
-
-        &-link {
+        font-weight: bold;
```
Wardormeur: Wont' that look very... aggressive ? I mean, there is 6 news displayed next to another.
Could we get a screenshot ? :)

src/dashboard/cd-dashboard-news.vue
```diff
@@ -114,36 +112,15 @@
...
-        flex-direction: column;
-        margin: 0 16px 0 16px;
-        max-width: 30%;
+        max-width: 75px;
       }
 
       &-right{
         margin: 0 16px 0 16px;
-        max-width: 70%;
+        max-width: 425px;
```
Wardormeur: What happens when it's over 425 (ie responsive)?

src/dashboard/cd-dashboard-news.vue
heyitsmaimai: i'll put one up on the card in a minute

src/dashboard/cd-dashboard-news.vue
heyitsmaimai: it wraps. but also need to fix mobile padding for the date

Wardormeur: Fix it and it's a go :)

heyitsmaimai: it's fixed!!!

Wardormeur:
I believe in you :<

PR #200, opened 2018-09-11
Add tests for stats

Wardormeur:
and lack of stats

### DIFF
--- a/cypress/integration/dashboard/stats_spec.js
+++ b/cypress/integration/dashboard/stats_spec.js
@@ -1,0 +1,52 @@
+import homePage from '../../pages/home';
+
+describe('Homepage stats', () => {
+  beforeEach(() => {
+    cy.server();
+    cy.route('POST', '/api/2.0/dojos/users', [{ userTypes: ['champion'] }]).as('userDojos');
+  });
+
+  it('should show the stats component with its default category: Youth', () => {
+    cy.route('/api/2.0/users/instance', 'fx:parentLoggedIn').as('loggedIn');
+    cy.visit('/home');
+    cy.wait('@loggedIn');
+    cy.get(homePage.statsTitle).invoke('text').should('eq', 'Dojo stats');
+    cy.get(homePage.statsCategory).invoke('text').should('eq', 'Youth');
+  });
+
+  it('should show a message pushing champions to use Zen when no stats is available', () => {
+    cy.route('/api/2.0/users/instance', 'fx:parentLoggedIn').as('loggedIn');
+    cy.route('POST', '/api/2.0/dojos/load-dojo-users', { response: [] }).as('dojoUsers');
+    cy.visit('/home');
+    cy.wait('@loggedIn');
+    cy.wait('@userDojos');
+    cy.wait('@dojoUsers');
+    cy.get(homePage.statsUseZenMessage).invoke('text').should('eq', "No statistics are available at the moment.\n        The more Zen is used, the more you\'ll find about your Dojo!");
+  });
+  describe('Youth', () => {
+    it('should show the stats for the number of children booked', () => {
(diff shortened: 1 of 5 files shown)


cypress/integration/dashboard/stats_spec.js
```diff
@@ -0,0 +1,52 @@
+import homePage from '../../pages/home';
+
+describe('Homepage stats', () => {
+  beforeEach(() => {
+    cy.server();
+    cy.route('POST', '/api/2.0/dojos/users', [{ userTypes: ['champion'] }]).as('userDojos');
```
DanielBrierton: Should include permissions? Stats would be visible to mentors with dojo-admin rights, wouldn't they?

Wardormeur: ticketing admin rather ?

DanielBrierton: Only Dojo admin gives you access to the users of a Dojo though, right?

Wardormeur: Arrgh, you're right. So we can bin the mentors from that pie chart cc @conoramurphy 
I guess at that point, it's more sensible to just say champions @DanielBrierton  ? If we agree the perms are not editable anymore

DanielBrierton: Well, it's one thing to say perms aren't editable anymore (also, not entirely sure I agree just yet since I don't quite know what would replace it), but they are... And probably will be for a while...

So I'd say for now, this needs to check for dojo-admin until we actually make that decision and make the necessary perms changes for it.

cypress/integration/dashboard/stats_spec.js
```diff
@@ -17,9 +17,12 @@ describe('Homepage stats', () => {
   it('should show a message pushing champions to use Zen when no stats is available', () => {
     cy.route('/api/2.0/users/instance', 'fx:parentLoggedIn').as('loggedIn');
     cy.route('POST', '/api/2.0/dojos/load-dojo-users', { response: [] }).as('dojoUsers');
+    cy.route('POST', '/api/2.0/events/applications/search', []).as('bookedChildren');
     cy.visit('/home');
     cy.wait('@loggedIn');
     cy.wait('@userDojos');
+    cy.wait('@bookedChildren');
+    cy.wait('@dojoUsers');
```
DanielBrierton: Waiting for dojoUsers twice? Is this one meant to be userDojos like below?

Wardormeur: Technically, it's not wrong.
There are 2 calls to @dojoUsers : one for o13 and one for u13

Wardormeur:
Cypress is failing until the strings (https://github.com/CoderDojo/cp-translations/pull/89) are being merged

cypress/integration/dashboard/stats_spec.js
conoramurphy: Sorry just to clarify. Are we saying it's extra work to show it to mentors with dojo-admin rights?

conoramurphy: Or that we can't seperate the mentors easily? Why would we leave out mentors.

Wardormeur: Because either we rely on perms, which we dont' want to for deprecation purpose, or we rely on role, in which case a mentor doesn't have dojo-admin by default, which is the current restriction for the API call letting us loading all users of a dojos.
Ticketing-admin <perm> can get this data, but only event by event, for users which are part of the events.

Wardormeur: tldr : the API is not done for stats purpose

Wardormeur: cc @conoramurphy (please tell me if I'm wrong @DanielBrierton )

PR #203, opened 2018-09-12
define title and subtitle styling

heyitsmaimai:


Wardormeur:
Could you add a "default margin" as well ?
I think there should be more than title and sub : in the current design, we have h1, h2, and h4 (...) altogether on the page. Links as well are using a different styling that c/should (?) be ported over

Wardormeur:
Todo during last review : apply those styles to corresponding areas

PR #208, opened 2018-09-13
Dashboard/stats tracking navigation

Wardormeur:
Actual commit :   https://github.com/CoderDojo/cp-zen-frontend/commit/2da70878f48d96162d9ad4d1004dc6c887a6794c


src/dashboard/cd-dashboard.vue
```diff
@@ -7,34 +7,77 @@
...
+        return this.filterUserDojos('champion').length ? 'champion' :
+          this.filterUserDojos('mentor').length ? 'mentor' :
+            this.filterUserDojos('parent-guardian').length ? 'parent-guardian' :
+              'ninja';
+        /* eslint-enable no-nested-ternary */
+      },
+    },
+    methods: {
+      filterUserDojos(role) {
+        return this.userDojos.filter(ud => ud.userTypes.includes(role));
```
DanielBrierton: `includes` isn't supported in IE11 if we still care...

src/common/directives/cd-ga-track-exit-nav.js
```diff
@@ -0,0 +1,14 @@
+import Vue from 'vue';
+
+export default {
+  bind(el) {
+    el.addEventListener('click', (e) => {
+      e.preventDefault();
```
DanielBrierton: This breaks middle/ctrl click, right?

Wardormeur: It does, indeed. I can not prevent the event, but I'm not sure how to interpret the sendBeacon from MDN. From the content, it should carry over, but I'm expecting ga to polyfill it with xhr for IE11 (https://caniuse.com/#feat=beacon), which requires then the callback

src/dashboard/cd-dashboard.vue
Wardormeur: 5% of traffic from IE 11 :/
Won't babel transform it into something usable, like any es6 function ?

DanielBrierton: Babel only transforms syntax, not methods. You'd need to polyfill, unfortunately. Or implement it in a different way.

Wardormeur: I mean, filter isn't supported either then so.. well.. too late ? :dagger:

DanielBrierton: Filter is supported since IE9... https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter#Browser_compatibility

Wardormeur: whoops, typed filter in caniuse and it's the CSS attribute, not the js one :(
Apparently we already use includes in FTB, so.. Well, farewell IE11

DanielBrierton: I did the same thing originally. 😅

Or we remove the usage of `includes` in FTB..? 🤔 Should probably do a general test of IE11 sometime soon...

PR #212, opened 2018-09-17
Hide sidebar if nothing is to be displayed

Wardormeur:


src/dashboard/cd-dashboard.vue
```diff
@@ -6,7 +6,7 @@
         <dashboard-projects/>
         <dashboard-news />
       </div>
-      <div class="cd-dashboard__right-column">
+      <div class="cd-dashboard__right-column" v-if="userProfile && (childrenAreVisible || statsAreVisible)">
```
DanielBrierton: Show by default, hide when we know there's nothing to show

PR #226, opened 2018-09-20
Set default browser to chrome as electron fails randomly

Wardormeur:


Wardormeur:
@DanielBrierton it passed \o

PR #230, opened 2018-10-17
During booking, add kids in sequence to prevent orphaning

DanielBrierton:


test/unit/specs/events/order/cd-event-sessions.spec.js
```diff
@@ -449,17 +449,36 @@ describe('Event sessions component', () => {
       });
     });
 
-    describe('methods.addNewChildren()', () => {
+    describe.only('methods.addNewChildren()', () => {
```
Wardormeur: WILD

DanielBrierton: You could've used the new suggestion feature 😛 

```suggestion
    describe('methods.addNewChildren()', () => {
```

PR #236, opened 2019-01-18
CDF delete user

Wardormeur:
WIP.


src/users/cdf-manage.vue
```diff
@@ -0,0 +1,236 @@
...
+        this.user = {};
+        this.children = [];
+        this.dojos = [];
+        this.memberships = [];
+        this.deleted = false;
+      },
+    },
+    async created() {
+      this.reset();
+      if (!this.isLoggedIn) return this.$router.replace('/cdf');
```
DanielBrierton: Shouldn't be needed with the nav guard, right?

Wardormeur: Very likely a relict of before I implemented the CDF navguard, yes
Good catch :+1:

PR #248, opened 2019-03-05
Feature/dojo events ics

Wardormeur:
Add link to the ics for all upcoming events of a dojo


Wardormeur:
Ref: https://github.com/CoderDojo/community-platform/issues/1246

Wardormeur:
screenshot from 2019-03-05 00-54-15

Wardormeur:
Note: "[Beta]" has been removed
Translation will be done when it's approved to avoid conficts

Wardormeur:
Random thoughts, but it might be worth checking how this behave with private event. We might ensure we don't pull the private events if the user is not part of the Dojo

conoramurphy:
So I'll ask some stupid questions...

1) This is tracking all future events into the calendar correct? Then I think it should say "Add all events to your calendar" or else it would look like you might be adding only the one viewable right now
2) Can we track clicks? (not expecting a crazy amount just curious)
3) I think we should only show this if you are member of the Dojo. Does that make sense?

Wardormeur:
> 1. This is tracking all future events into the calendar correct? Then I think it should say "Add all events to your calendar" or else it would look like you might be adding only the one viewable right now
Well, the viewable events right now are all future events for a Dojo. How the ICS is afterwards treated is up to the application. For example, Google calendar refreshes ICS by URL every "few hours" (https://webapps.stackexchange.com/questions/6313/how-often-does-google-calendar-update-its-other-calendar-feeds)

> 2. Can we track clicks? (not expecting a crazy amount just curious)
Should be doable, yes.
> 3. I think we should only show this if you are member of the Dojo. Does that make sense?
We could, but I'm unsure why we would. First that's hiding it for no reason (it's part of the API and we already show public events to anybody), and then there is very few reasons to come back to the Dojo page once you're joined, as the event link is sent by emails.

As a 4th, there was consideration in:
 - The link currently makes you download the ICS file. However, downloading it will not give you updates as the link needs to be imported as the source, rather than the calendar being downloaded and then imported. The reason for that is that the URL is what gives the calendar updates. Updating with a file will not automatically update the calendar.
I would recommend to transform the link into an input containing the url, with a "copy" button. As well I would add a button to "Open" the url, using the `webcal` protocol, for users having a preconfigured calendar application.

conoramurphy:
1. Fair enough, your text is fine then
3. It would get passed all the "showing private events" stuff. But if that's too big an action to fix a smaller concern I'm ok with how it is.

> I would recommend transforming the link into an input containing the url, with a "copy" button. As well I would add a button to "Open" the url, using the webcal protocol, for users having a preconfigured calendar application.

I think that input mechanism is a bit weird to me with an open and a close button. If we want to do a url just that's fine. However at that point I would just put the url in the sidebar with a title describing it (Calendar Link) and let people do with it what they will. I think an input wiht 2 buttons is strong overkill here

Wardormeur:
Click on the link : 
Screenshot from 2019-04-01 15-47-01
Replace the content with the url with the capacity to copy the link
Screenshot from 2019-04-01 15-48-40

and the option to open it with the `webdav` protocol so that local calendar (such as iCal or gcalendar on android) can open it, either via
 - a dropdown with a single option, called "Open with your calendar"
 - or directly as a secondary button next to copy with a single "external link" icon

conoramurphy:
Ah yes that makes more sense to me now. I still think for the complexity it might be easier in the left-hand column, particularly for responsiveness.

Wardormeur:
As a sidebar element
Screenshot from 2019-04-01 16-39-29

Wardormeur:
As part of the sidebar, in the time area
Screenshot from 2019-04-01 16-47-19

conoramurphy:
I'm fine with the last one if you are?

PR #251, opened 2019-04-01
Add new page for accept/deny access to Dojo

Wardormeur:
through requests to join


Wardormeur:
Screenshot from 2019-04-01 15-27-09

PR #255, opened 2019-04-23
Add user name to redirection link on duplicate user

Wardormeur:
To ensure they don't look further but the page they're on when the user need to be modified


Wardormeur:
Requires https://github.com/CoderDojo/cp-zen-platform/pull/1362

PR #261, opened 2019-06-10
Ensure en_US locale returns projects

ArayB:
When the users cookie locale is en_US, or presumably any other en_
locale we were sending the entire locale to learning admin to get
projects back.
Learning admin expects the locale to be 'en' for english.

This PR checks the locale after retrieving it from the user's cookies and formats it if needed.

If we send the incorrect locale the projects on the users dashboard don't load:

Screenshot 2019-06-10 at 13 39 49

The locales as defined in projects admin are:
```
two_letter_locales = %i[
  en
]

four_letter_locales = %i[
  ar-SA az-AZ bg-BG bn-BD ca-ES cs-CZ cy-GB da-DK de-DE el-GR es-ES es-LA et-EE fa-IR fi-FI fr-CA
  fr-FR he-IL hi-IN hr-HR hu-HU id-ID it-IT ja-JP ko-KR me-ME ml-IN nl-NL no-NO pl-PL pt-BR pt-PT
  ro-RO ru-RU sk-SK sl-SI sr-SP sv-SE tr-TR uk-UA vi-VN vls-BE zh-CN zh-TW
]
```

So en is the only two letter case.

It turned out that there was already a check on the `created` vue event to get the `en` locale as a default if no projects had been retrieved.
This wasn't working because the failing request wasn't being handled and an error was being thrown.

I've amended the request to the projects site to swallow an error and let the fallback occur.

### DIFF
--- a/src/dashboard/cd-dashboard-projects.vue
+++ b/src/dashboard/cd-dashboard-projects.vue
@@ -4,7 +4,7 @@
     <h2 class="cd-dashboard-projects__header visible-xs">{{ $t('Here are some projects you can try') }}</h2>
     <hr class ="cd-dashboard-projects__divider visible-xs">
     <div v-show="isDisplayable" class="cd-dashboard-projects__cards">
-      <a class="cd-dashboard-projects__card" v-for="project in projects" :key="project.id" 
+      <a class="cd-dashboard-projects__card" v-for="project in projects" :key="project.id"
         :href="`https://projects.raspberrypi.org/${locale}/projects/${project.attributes.repositoryName}`"
         v-ga-track-exit-nav>
         <img :src="project.attributes.content.heroImage" />
@@ -41,7 +41,11 @@
       userLocale() {
         const langCookie = LocaleService.getUserLocale();
         if (langCookie) {
-          return langCookie.replace(/"/g, '').replace('_', '-');
+          let formattedLocale = langCookie.replace(/"/g, '').replace('_', '-');
+          if (/en-.*/.test(formattedLocale)) {
+            formattedLocale = 'en';
+          }
+          return formattedLocale;
         }
         return 'en';
       },
--- a/src/dashboard/cd-dashboard-projects.vue
+++ b/src/dashboard/cd-dashboard-projects.vue
@@ -52,17 +52,15 @@
     },
     methods: {
       async loadProjects(locale = 'en') {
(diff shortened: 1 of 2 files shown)


src/dashboard/cd-dashboard-projects.vue
```diff
@@ -41,7 +41,11 @@
       userLocale() {
         const langCookie = LocaleService.getUserLocale();
         if (langCookie) {
-          return langCookie.replace(/"/g, '').replace('_', '-');
+          let formattedLocale = langCookie.replace(/"/g, '').replace('_', '-');
+          if (/en-.*/.test(formattedLocale)) {
```
DanielBrierton: I reckon this would happen for other locales, like es-ES, es-AR, es-MX, etc. Worth checking, and if so probably makes more sense to just pull out the first two characters, or split by the `-` (or the `_` and don’t do the replacement...)

ArayB: Yes good point.
Will check on that.

src/dashboard/cd-dashboard-projects.vue
```diff
@@ -41,17 +41,28 @@
...
         return 'en';
       },
     },
     methods: {
       async loadProjects(locale = 'en') {
-        const projects = (await ProjectsService.list(locale, { order: 'desc' })).body;
-        if (projects) {
-          this.projects = projects.data.slice(0, 3);
-        }
+        await ProjectsService.list(locale, { order: 'desc' })
```
DanielBrierton: You should be able to just wrap the existing code in try/catch. It works around async/await.

```javascript
try {
  const projects = (await ProjectsService.list(locale, { order: 'desc' })).body;
  if (projects) {
    this.projects = projects.data.slice(0, 3);
  }
} catch() {}
```

Makes it a little cleaner and easier to see what's going on.

src/dashboard/cd-dashboard-projects.vue
DanielBrierton: I guess there's a question about whether this is needed anymore. It'll result in one less request, and it's unlikely we're gonna differentiate between en-US and en-GB, so I guess keep it..?

ArayB: I was thinking that as well, it feels like it's probably worth it to save a request so will leave it in.

(Fast forward to bug in three years time when en-GB and en-US are added to projects api)

PR #312, opened 2020-07-23
POC: Text direction component

glenpike:
Add a component that can be used to wrap others and set the text direction.

Text direction can be set as a prop on the component which will override everything else, otherwise, if the config for a chosen language specifies a 'dir' property, this will be used.  Default is 'ltr'

Currently added as a wrapper around the whole app, then a specific wrapper around the Covid banner to set it to LTR as there are no translations for this currently.

May need using in other parts of the site - we will need feedback on where to put it I expect.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -1,11 +1,13 @@
 <template>
   <div id="app">
-    <cd-header></cd-header>
-    <div class="cd-menu__content-container">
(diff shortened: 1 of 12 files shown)


glenpike:
What the find page looks like with rtl set:

Screenshot 2020-07-23 at 12 52 57

PR #320, opened 2020-10-12
Display online status

glenpike:
If a club has ticked the 'Online' box, show in the club details:

Screenshot 2020-10-12 at 16 25 26

Also show in the search results:

Screenshot 2020-10-12 at 16 54 15
Screenshot 2020-10-12 at 16 54 30


glenpike:
Looked at how the tests work and not sure it's going to be easy.  It seems that the tests rely on existing clubs (they search for 'dublin' then pick specific ones from the list) - so we'd have to setup an existing club with the online sessions field ticked.  This also implies that tests are run against a production DB...  There do not seem to be any tests for editing a Dojo.

patch0 (approved):

PR #342, opened 2021-12-23
Update google maps API key for test/dev envs.

patch0:
Cypress tests were failing, because the cookie banner test couldn't geolocate "dublin".  I found this by running Cypress locally and opening the dev tools inside the chrome session, where there was a warning message about the expired API key.


dumbledad (approved):
LGTM, I ♥️ small PRs!

